PR #1, opened 2021-01-05
Implement websocket server and client.

arijitAD:
- [x] Implement Web socket server.
- [ ] Implement Web socket client.

### DIFF
--- a/src/ws.nim
+++ b/src/ws.nim
@@ -1,0 +1,570 @@
+import chronos, chronicles, httputils, strutils, base64, std/sha1, random,
+    streams, nativesockets, uri, times, chronos.timer, tables
+
+const
+  MaxHttpHeadersSize = 8192        # maximum size of HTTP headers in octets
+  MaxHttpRequestSize = 128 * 1024  # maximum size of HTTP body in octets
+  HttpHeadersTimeout = timer.seconds(120) # timeout for receiving headers (120 sec)
+  HttpBodyTimeout = timer.seconds(12)     # timeout for receiving body (12 sec)
+  HeadersMark = @[byte(0x0D), byte(0x0A), byte(0x0D), byte(0x0A)]
+  WebsocketUserAgent* = "nim-ws (https://github.com/status-im/nim-ws)"
+
+type
+  ReadyState* = enum
+    Connecting = 0 # The connection is not yet open.
+    Open = 1       # The connection is open and ready to communicate.
+    Closing = 2    # The connection is in the process of closing.
+    Closed = 3     # The connection is closed or couldn't be opened.
+
+  WebSocket* = ref object
+    tcpSocket*: StreamTransport
+    version*: int
+    key*: string
+    protocol*: string
+    readyState*: ReadyState
+    masked*: bool # send masked packets
+    header*: HttpHeaders
+
+  AsyncCallback = proc (transp: StreamTransport,
+      header: HttpRequestHeader): Future[void] {.closure, gcsafe.}
+  HttpServer* = ref object of StreamServer
+    callback: AsyncCallback
+
+  ReqStatus = enum
+    Success, Error, ErrorFailure
+
+  WebSocketError* = object of IOError
+
+  HttpHeaders* = ref object
+    table*: TableRef[string, seq[string]]
+
+type
+  AsyncHttpClient* = ref object
+    connected: bool
+    currentURL: Url          ## Where we are currently connected.
+    headers: seq[HttpHeader] ## Headers to send in requests.
+    userAgent: string
+    contentTotal: BiggestInt
+    contentProgress: BiggestInt
+    oneSecondProgress: BiggestInt
+    transp: StreamTransport
+    buf: seq[byte]
+
+template `[]`(value: uint8, index: int): bool =
+  ## Get bits from uint8, uint8[2] gets 2nd bit.
+  (value and (1 shl (7 - index))) != 0
+
+proc nibbleFromChar(c: char): int =
+  ## Converts hex chars like `0` to 0 and `F` to 15.
+  case c:
+    of '0'..'9': (ord(c) - ord('0'))
+    of 'a'..'f': (ord(c) - ord('a') + 10)
+    of 'A'..'F': (ord(c) - ord('A') + 10)
+    else: 255
+
+proc nibbleToChar(value: int): char =
+  ## Converts number like 0 to `0` and 15 to `fg`.
+  case value:
+    of 0..9: char(value + ord('0'))
+    else: char(value + ord('a') - 10)
+
+proc decodeBase16*(str: string): string =
+  ## Base16 decode a string.
+  result = newString(str.len div 2)
+  for i in 0 ..< result.len:
+    result[i] = chr(
+      (nibbleFromChar(str[2 * i]) shl 4) or
+      nibbleFromChar(str[2 * i + 1]))
+
+proc encodeBase16*(str: string): string =
+  ## Base61 encode a string.
+  result = newString(str.len * 2)
+  for i, c in str:
+    result[i * 2] = nibbleToChar(ord(c) shr 4)
+    result[i * 2 + 1] = nibbleToChar(ord(c) and 0x0f)
+
+proc genMaskKey(): array[4, char] =
+  ## Generates a random key of 4 random chars.
+  proc r(): char = char(rand(255))
+  [r(), r(), r(), r()]
+
+proc handshake*(ws: WebSocket, header: HttpRequestHeader) {.async.} =
+  ## Handles the websocket handshake.
+  ws.version = parseInt(header["Sec-WebSocket-Version"])
+  ws.key = header["Sec-WebSocket-Key"].strip()
+  if header.contains("Sec-WebSocket-Protocol"):
+    let wantProtocol = header["Sec-WebSocket-Protocol"].strip()
+    if ws.protocol != wantProtocol:
+      raise newException(WebSocketError,
+        "Protocol mismatch (expected: " & ws.protocol & ", got: " &
+        wantProtocol & ")")
+
+  let
+    sh = secureHash(ws.key & "258EAFA5-E914-47DA-95CA-C5AB0DC85B11")
+    acceptKey = base64.encode(decodeBase16($sh))
+
+  var response = "HTTP/1.1 101 Web Socket Protocol Handshake\c\L"
+  response.add("Sec-WebSocket-Accept: " & acceptKey & "\c\L")
+  response.add("Connection: Upgrade\c\L")
+  response.add("Upgrade: webSocket\c\L")
+
+  if ws.protocol != "":
+    response.add("Sec-WebSocket-Protocol: " & ws.protocol & "\c\L")
+  response.add "\c\L"
+
+  discard await ws.tcpSocket.write(response)
+  ws.readyState = Open
+
+proc newWebSocket*(header: HttpRequestHeader, transp: StreamTransport,
+                    protocol: string = ""): Future[WebSocket] {.async.} =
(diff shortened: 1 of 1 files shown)


src/ws.nim
```diff
@@ -0,0 +1,570 @@
...
+proc encodeBase16*(str: string): string =
+  ## Base61 encode a string.
+  result = newString(str.len * 2)
+  for i, c in str:
+    result[i * 2] = nibbleToChar(ord(c) shr 4)
+    result[i * 2 + 1] = nibbleToChar(ord(c) and 0x0f)
+
+proc genMaskKey(): array[4, char] =
+  ## Generates a random key of 4 random chars.
+  proc r(): char = char(rand(255))
```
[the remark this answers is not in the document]
dryajov: We agreed on using a similar style to the rest of our stack, which is pass an external randomness source.

src/ws.nim
```diff
@@ -0,0 +1,570 @@
...
+    rsv3: bool ## MUST be 0
+    opcode: Opcode ## Defines the interpretation of the "Payload data".
+    mask: bool ## Defines whether the "Payload data" is masked.
+    data: string ## Payload data
+
+proc encodeFrame(f: Frame): string =
+  ## Encodes a frame into a string buffer.
+  ## See https://tools.ietf.org/html/rfc6455#section-5.2
+
+  var ret = newStringStream()
```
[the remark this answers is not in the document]
dryajov: lets avoid using `result` altogether since it's been both a source of countless arguments and an apparent source of bugs as well.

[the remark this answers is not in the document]
dryajov: Agree on avoiding the use of `StringStream`.

src/ws.nim
```diff
@@ -0,0 +1,570 @@
...
+    raise newException(WebSocketError, "Failed to send data: " &
+        getCurrentExceptionMsg())
+
+proc close*(ws: WebSocket) =
+  ## Close the Socket, sends close packet.
+  ws.readyState = Closed
+  proc close() {.async.} =
+    await ws.send("", Close)
+    ws.tcpSocket.close()
+  asyncCheck close()
```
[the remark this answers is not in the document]
dryajov: What is the intent of using the `asyncCheck` here @arijitAD?

src/ws.nim
```diff
@@ -0,0 +1,570 @@
...
+proc decodeBase16*(str: string): string =
+  ## Base16 decode a string.
+  result = newString(str.len div 2)
+  for i in 0 ..< result.len:
+    result[i] = chr(
+      (nibbleFromChar(str[2 * i]) shl 4) or
+      nibbleFromChar(str[2 * i + 1]))
+
+proc encodeBase16*(str: string): string =
+  ## Base61 encode a string.
```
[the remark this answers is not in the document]
dryajov: Like Zah mentioned, lets avoid duplicating this helpers as most/all are available in `nim-stew`.

src/ws.nim
```diff
@@ -0,0 +1,570 @@
...
+
+proc toString(bytes: seq[byte]): string =
+  result = ""
+  for byte in bytes: result.add(char(byte))
+
+proc receiveFrame(ws: WebSocket): Future[Frame] {.async.} =
+  ## Gets a frame from the WebSocket.
+  ## See https://tools.ietf.org/html/rfc6455#section-5.2
+
+  if cast[int](ws.tcpSocket.fd) == -1:
```
dryajov, line 279: I think this is a bit unsafe and `fd` should be private. In general, I don't believe we check for `closed` when using chronos sockets and instead rely on exceptions.  

For example instead of the check just handle the `TransportUseClosedError` and set the flag in the `except` clause. I agree that this is cumbersome and we should perhaps add a `closed/isClosed` method to chronos @cheatfate?

src/ws.nim
```diff
@@ -0,0 +1,570 @@
...
+  # If there are more parts read and wait for them
+  while frame.fin != true:
+    frame = await ws.receiveFrame()
+    if frame.opcode != Cont:
+      raise newException(WebSocketError, "Socket closed")
+    result[1].add frame.data
+  return
+
+proc receiveStrPacket*(ws: WebSocket): Future[string] {.async.} =
+  ## Wait only for a string packet to come. Errors out on Binary packets.
```
dryajov, line 373: this is a bit confusing, the comment says that this should only handle string packets but it also responds to pings?

src/ws.nim
```diff
@@ -0,0 +1,570 @@
...
+  if header.len != 2:
+    ws.readyState = Closed
+    raise newException(WebSocketError, "Socket closed")
+
+  let b0 = header[0].uint8
+  let b1 = header[1].uint8
+
+  # Read the flags and fin from the header.
+  result.fin = b0[0]
+  result.rsv1 = b0[1]
```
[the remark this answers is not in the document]
arnetheduck: hm, wouldn't a `set` be more natural in nim? https://nim-lang.org/docs/manual.html#types-set-type

I generally wouldn't trust the implementation of `bitsize` to work well / portably across backends/compilers, it's rarely used except to support FFI.

src/ws.nim
```diff
@@ -0,0 +1,570 @@
...
+  HttpServer* = ref object of StreamServer
+    callback: AsyncCallback
+
+  ReqStatus = enum
+    Success, Error, ErrorFailure
+
+  WebSocketError* = object of IOError
+
+  HttpHeaders* = ref object
+    table*: TableRef[string, seq[string]]
```
arnetheduck, line 39: just `Table` - `ref` loses mutability protections in the language and is susceptible to `nil` issues.

src/ws.nim
```diff
@@ -0,0 +1,570 @@
...
+      header: HttpRequestHeader): Future[void] {.closure, gcsafe.}
+  HttpServer* = ref object of StreamServer
+    callback: AsyncCallback
+
+  ReqStatus = enum
+    Success, Error, ErrorFailure
+
+  WebSocketError* = object of IOError
+
+  HttpHeaders* = ref object
```
arnetheduck, line 38: Also just `object` - `ref object` is a shortcut that prevents the non-ref use of the type  - the caller can always make a `ref HttpHeaders` if their use case demands it, but not the other way around - `ref object` should only be used if the semantic of the type requires it (mainly when managing a resource that can't be copied) - in particular, using `ref HttpHeaders` explicitly makes it easier to track down the less safe code that is affected by `ref` semantics.

src/ws.nim
```diff
@@ -0,0 +1,570 @@
...
+    result[i * 2 + 1] = nibbleToChar(ord(c) and 0x0f)
+
+proc genMaskKey(): array[4, char] =
+  ## Generates a random key of 4 random chars.
+  proc r(): char = char(rand(255))
+  [r(), r(), r(), r()]
+
+proc handshake*(ws: WebSocket, header: HttpRequestHeader) {.async.} =
+  ## Handles the websocket handshake.
+  ws.version = parseInt(header["Sec-WebSocket-Version"])
```
cheatfate, line 93: std.parseInt() should not be used, because it raises too many exceptions and allows negative values like `-1`.

src/ws.nim
```diff
@@ -0,0 +1,570 @@
...
+  # Payload length can be 7 bits, 7+16 bits, or 7+64 bits.
+  var finalLen: uint = 0
+
+  let headerLen = uint(b1 and 0x7f)
+  if headerLen == 0x7e:
+    # Length must be 7+16 bits.
+    var length = await ws.tcpSocket.read(2)
+    if length.len != 2:
+      raise newException(WebSocketError, "Socket closed")
+    finalLen = cast[ptr uint16](length[0].addr)[].htons
```
cheatfate, line 318: you can replace `nativesockets.htons` with `stew` primitives.

src/ws.nim
```diff
@@ -0,0 +1,570 @@
...
+    maskKey = await ws.tcpSocket.read(4)
+    if maskKey.len != 4:
+      raise newException(WebSocketError, "Socket closed")
+
+  # Read the data.
+  var data: seq[byte]
+  data = await ws.tcpSocket.read(int finalLen)
+  result.data = toString(data)
+  if result.data.len != int finalLen:
+    raise newException(WebSocketError, "Socket closed")
```
cheatfate, line 351: It could confuse everybody if because of incorrect length you will receive `Socket closed` error, And socket is not even closed at this moment.

src/ws.nim
```diff
@@ -0,0 +1,570 @@
...
+proc receivePacket*(ws: WebSocket): Future[(Opcode, string)] {.async.} =
+  ## Wait for a string or binary packet to come in.
+  var frame = await ws.receiveFrame()
+  result[0] = frame.opcode
+  result[1] = frame.data
+  # If there are more parts read and wait for them
+  while frame.fin != true:
+    frame = await ws.receiveFrame()
+    if frame.opcode != Cont:
+      raise newException(WebSocketError, "Socket closed")
```
cheatfate, line 368: Same confusing error message as previous one.

src/ws.nim
```diff
@@ -0,0 +1,570 @@
...
+  if header.meth notin {MethodGet}:
+    # Request method is either PUT or DELETE.
+    debug "GET method is only allowed", address = transp.remoteAddress()
+    if await transp.sendHTTPResponse(header.version, Http405):
+      result = Error
+    else:
+      result = ErrorFailure
+    return
+
+  if header.contentLength() > MaxHttpRequestSize:
```
cheatfate, line 423: `header.contentLength()` could return `0` and `-1` and pass all the checks here.
https://github.com/status-im/nim-http-utils/blob/master/httputils.nim#L632-L635

src/ws.nim
```diff
@@ -0,0 +1,570 @@
...
+
+  let b0 = header[0].uint8
+  let b1 = header[1].uint8
+
+  # Read the flags and fin from the header.
+  result.fin = b0[0]
+  result.rsv1 = b0[1]
+  result.rsv2 = b0[2]
+  result.rsv3 = b0[3]
+  result.opcode = (b0 and 0x0f).Opcode
```
cheatfate, line 302: `Opcode` is enum with holes, such conversion could lead to exception if `b0` is in `[3, 7]` or `[b, f]`.

src/ws.nim
```diff
@@ -0,0 +1,570 @@
...
+  if headerLen == 0x7e:
+    # Length must be 7+16 bits.
+    var length = await ws.tcpSocket.read(2)
+    if length.len != 2:
+      raise newException(WebSocketError, "Socket closed")
+    finalLen = cast[ptr uint16](length[0].addr)[].htons
+
+  elif headerLen == 0x7f:
+    # Length must be 7+64 bits.
+    var length = await ws.tcpSocket.read(8)
```
cheatfate, line 322: If you need exactly 8 bytes to read, you need to use `readExactly` call.

src/ws.nim
```diff
@@ -0,0 +1,570 @@
...
+
+  if ws.masked == result.mask:
+    # Server sends unmasked but accepts only masked.
+    # Client sends masked but accepts only unmasked.
+    raise newException(WebSocketError, "Socket mask mismatch")
+
+  var maskKey: seq[byte]
+  if result.mask:
+    # Read the mask.
+    maskKey = await ws.tcpSocket.read(4)
```
cheatfate, line 342: `readExactly`?

src/ws.nim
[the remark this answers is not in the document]
arijitAD: I was planning to use discard. But then I read in the docs that `asyncCheck` should be used instead of `discard`.
https://github.com/nim-lang/Nim/blob/version-1-4/lib/pure/asyncfutures.nim#L408

arijitAD:
Closing this. Addressed all comments #2

PR #2, opened 2021-01-12
[WIP] Web socket client implementation.

arijitAD:
- [x] Implement simple HTTP client.
- [x] Implement Websocket client.
- [x] Implement Echo Server.
- [x] Validate Websocket client request.
- [x] Add tests.


arnetheduck:
fyi we don't use `nimpretty` generally: it's broken and buggy - in particular, can break correct code and doesn't do cost-based formatting meaning the output is often ugly / hard to read

src/ws.nim
```diff
@@ -1,13 +1,13 @@
...
 const
-  MaxHttpHeadersSize = 8192        # maximum size of HTTP headers in octets
-  MaxHttpRequestSize = 128 * 1024  # maximum size of HTTP body in octets
+  MaxHttpHeadersSize = 8192       # maximum size of HTTP headers in octets
+  MaxHttpRequestSize = 128 * 1024 # maximum size of HTTP body in octets
   HttpHeadersTimeout = timer.seconds(120) # timeout for receiving headers (120 sec)
-  HttpBodyTimeout = timer.seconds(12)     # timeout for receiving body (12 sec)
-  HeadersMark = @[byte(0x0D), byte(0x0A), byte(0x0D), byte(0x0A)]
   WebsocketUserAgent* = "nim-ws (https://github.com/status-im/nim-ws)"
+  CRLF* = "\c\L"
```
arnetheduck, line 9:
```suggestion
  CRLF* = "\r\n"
```

arijitAD:
> fyi we don't use `nimpretty` generally: it's broken and buggy - in particular, can break correct code and doesn't do cost-based formatting meaning the output is often ugly / hard to read

Is there any other formatting tool? or any standard guidelines that we follow?

src/ws.nim
```diff
@@ -254,7 +259,7 @@ proc send*(ws: WebSocket, text: string, opcode = Opcode.Text): Future[void] {.as
       let data = frame[i ..< min(frame.len, i + maxSize)]
       discard await ws.tcpSocket.write(data)
       i += maxSize
-      await sleepAsync(1)
+      await stepsAsync(1)
```
arnetheduck, line 253: there should be no steps or awaits here - `await write` already does all the necessary waiting

dryajov: Yes, we've already pointed this out in #1. There might be some exceptions for this rule, when for example there is a chance of ending up in a tight synchronous loop that can starve the main event loop, as a sort of a `yield`, but those are rare and for the most part there is never a need for such a `yield`.

src/ws.nim
```diff
@@ -254,7 +250,7 @@ proc send*(ws: WebSocket, text: string, opcode = Opcode.Text): Future[void] {.as
       let data = frame[i ..< min(frame.len, i + maxSize)]
       discard await ws.tcpSocket.write(data)
```
dryajov, line 251: Write can still fail writing all the data, you should check if the bytes written match the data size, otherwise it's an error.

src/ws.nim
```diff
@@ -535,7 +529,126 @@ func newHttpHeaders*(keyValuePairs:
       else:
         result.table[key] = @[pair.val]
 
-proc newAsyncWebsocketClient*(uri: Uri, protocols: seq[string] = @[]): Future[WebSocket] {.async.} =
+proc generateHeaders(requestUrl: Uri, httpMethod: string,
+    headers: HttpHeaders): string =
+  # GET
+  let upperMethod = httpMethod.toUpperAscii()
+  result = upperMethod
```
dryajov, line 536: We should avoid using `result` completely. I know that it's use proliferates through the standard library and other popular nim projects, but as per our internal convention, we've determined that its use is dangerous and leads to all sorts of hard to track bugs. The easiest example to convey this is in the presence of exceptions. If an exception is thrown in the middle of a multi step operation like the one in this proc, `result` ends up in an inconsistent state, one can imagine all sorts of scenarios where the exception is ignored and the incomplete/inconsistent `result` continues being used. The better approach is to simply stick to explicit result variables and `return` statments.

src/ws.nim
```diff
@@ -535,7 +529,126 @@ func newHttpHeaders*(keyValuePairs:
...
+
+  client.transp = await connect(initTAddress(url.hostname, port))
+
+  # May be connected through proxy but remember actual URL being accessed
+  client.currentURL = url
+  client.connected = true
+
+proc validateWSClientHandshake*(transp: StreamTransport,
+                       header: HttpResponseHeader): bool =
+  if header.code != 101:
```
dryajov, line 576: Probably better to use enums/constants for this sort of values.

src/ws.nim
dryajov: This still needs to be changed to the @arnetheduck suggestion.

src/ws.nim
dryajov: Let's make the check and handle incomplete writes appropriately.

src/ws.nim
dryajov: this still needs to be resolved.

dryajov: ie, the `sleepAsync` needs to go

src/ws.nim
dryajov: Still using `result` here.

dryajov:
@arijitAD, this looks like its almost there. Can we address the review and add a bit more tests?

dryajov:
@cheatfate it would be good if you could give this one more pass.

dryajov (approved):

src/http.nim
```diff
@@ -0,0 +1,250 @@
...
+func toTitleCase(s: string): string =
+  var tcstr = newString(len(s))
+  var upper = true
+  for i in 0..len(s) - 1:
+    tcstr[i] = if upper: toUpperAscii(s[i]) else: toLowerAscii(s[i])
+    upper = s[i] == '-'
+  return tcstr
+
+func toCaseInsensitive*(headers: HttpHeaders, s: string): string {.inline.} =
+  return toTitleCase(s)
```
dryajov, line 222: Maybe just toLower/toUpper case?

dryajov:
> @arijitAD, this looks like its almost there. Can we address the review and add a bit more tests?

I didn't realize that the comments are already addressed, my bad. 

But we're still missing testing for the most part. For starters I would like to see (not necessarily in this PR):

- Connect, handshake and initiate a websocket conversation
- Send full packets over the wire
- Send partial/broken packets over the wire

I would also imagine that there are compliance frameworks that will be able to test websockets implementations for spec adherence, correctness and performance. We should think about using those as part of the testing in the short to mid term.

ws.nimble
```diff
@@ -4,12 +4,12 @@ author = "Status Research & Development GmbH"
 description = "WS protocol implementation"
 license = "MIT"
 
-requires "nim >= 1.2.6"
-requires "chronos >= 2.5.2 & < 3.0.0"
+requires "nim >= 1.4.0"
```
dryajov, line 7: we should stick with 1.2.6 for now, until we've adopted 1.4.x across the board.

dryajov (approved):

src/ws.nim
```diff
@@ -0,0 +1,415 @@
+import chronos, httputils, strutils, base64, std/sha1, random, http,
```
cheatfate, line 1: Its unclear what `random` is going to be used from standard library or from `src` folder.

src/ws.nim
```diff
@@ -0,0 +1,415 @@
...
+    rsv1
+    fin
+  HeaderFlags = set[HeaderFlag]
+
+  HttpCode* = enum
+    Http101 = 101 # Switching Protocols
+
+proc handshake*(ws: WebSocket, header: HttpRequestHeader) {.async.} =
+  ## Handles the websocket handshake.
+  try: ws.version = parseInt(header["Sec-WebSocket-Version"])
```
cheatfate, line 42: Recently i have discovered new more appropriate procedure for int parsing. It will not raise exceptions anymore, but it still could return negative values.
https://github.com/nim-lang/Nim/blob/v1.2.6/lib/pure/parseutils.nim#L466

src/ws.nim
```diff
@@ -0,0 +1,415 @@
...
+    let wantProtocol = header["Sec-WebSocket-Protocol"].strip()
+    if ws.protocol != wantProtocol:
+      raise newException(WebSocketError,
+        "Protocol mismatch (expected: " & ws.protocol & ", got: " &
+        wantProtocol & ")")
+
+  var acceptKey: string
+  try:
+    let sh = secureHash(ws.key & "258EAFA5-E914-47DA-95CA-C5AB0DC85B11")
+    acceptKey = base64.encode(hexToByteArray[SHA1DigestSize]($sh))
```
cheatfate, line 61: It is better to use `stew` version of `base64` because it do not raise exceptions for `encode`.
https://github.com/status-im/nim-stew/blob/master/stew/base64.nim
And you can use `nimcrypto/sha1` which is also do not raise exceptions.

src/ws.nim
```diff
@@ -0,0 +1,415 @@
...
+  # Only need more bytes if data len is 7+16 bits, or 7+64 bits.
+  if f.data.len > 125 and f.data.len <= 0xffff:
+    # Data len is 7+16 bits.
+    var len = f.data.len.uint16
+    ret.add ((len shr 8) and 255).uint8
+    ret.add (len and 255).uint8
+  elif f.data.len > 0xffff:
+    # Data len is 7+64 bits.
+    var len = f.data.len
+    ret.add ((len shr 56) and 255).uint8
```
cheatfate, line 177: This looks like manual conversion to big-endian, i think `stew/endians2` could do it safer.

src/ws.nim
```diff
@@ -0,0 +1,415 @@
...
+    # Send stuff in 1 megabyte chunks to prevent IOErrors.
+    # This really large packets.
+    var i = 0
+    while i < frame.len:
+      let frameSize = min(frame.len, i + maxSize)
+      let res = await ws.tcpSocket.write(frame[i ..< frameSize])
+      if res != frameSize:
+        raise newException(ValueError, "Error while send websocket frame")
+      i += maxSize
+  except IOError, OSError, ValueError:
```
cheatfate, line 228: Why `IOError` is here?

src/ws.nim
```diff
@@ -0,0 +1,415 @@
...
+    # Wrap all exceptions in a WebSocketError so its easy to catch
+    raise newException(WebSocketError, "Failed to send data: " &
+        getCurrentExceptionMsg())
+
+proc sendStr*(ws: WebSocket, data: string, opcode = Opcode.Text): Future[
```
cheatfate, line 233: This could be represented like
```nim
proc sendStr*(ws: WebSocket, data: string, opcode = Opcode.Text): Future[void] =
  send(ws, toBytes(data), opcode)
```
Not that there no `{.async.}` pragma anymore, and procedure become just a proxy call for `send`.

src/ws.nim
```diff
@@ -0,0 +1,415 @@
...
+                void] {.async.} =
+  await send(ws, toBytes(data), opcode)
+
+proc close*(ws: WebSocket) =
+  ## Close the Socket, sends close packet.
+  ws.readyState = Closed
```
cheatfate, line 239: Please provide asynchronous version of `close` too. Usually in unit tests we can perform some sanity checks, right after `await close()` so we really need to know the moment when everything become actually closed.

src/ws.nim
```diff
@@ -0,0 +1,415 @@
...
+  ## See https://tools.ietf.org/html/rfc6455#section-5.2
+
+  # Grab the header.
+  var header = newSeq[byte](2)
+  try:
+    await ws.tcpSocket.readExactly(addr header[0], 2)
+  except TransportUseClosedError:
+    ws.readyState = Closed
+    raise newException(WebSocketError, "Socket closed")
+  except CatchableError:
```
cheatfate, line 256: By catching `CatchableError` you swallowing cancellation exceptions, so its better to avoid such construction or comment about reasons.

src/ws.nim
```diff
@@ -0,0 +1,415 @@
...
+    # Client sends masked but accepts only unmasked.
+    raise newException(WebSocketError, "Socket mask mismatch")
+
+  var maskKey = newSeq[byte](4)
+  if frame.mask:
+    # Read the mask.
+    await ws.tcpSocket.readExactly(addr maskKey[0], 4)
+
+  # Read the data.
+  var data = newSeq[byte](finalLen)
```
cheatfate, line 318: **SECURITY WARNING** I dont see any checks on `finalLen` size before allocating chunk of memory... So looks like its possible to allocate `high(uint64)` bytes of memory with frame which will be just 9 bytes size.

src/ws.nim
```diff
@@ -0,0 +1,415 @@
...
+  if header.code != ord(Http101):
+    raise newException(WebSocketError, "Server did not reply with a websocket upgrade: " &
+          "Header code: " & $header.code &
+          "Header reason: " & header.reason() &
+          "Address: " & $transp.remoteAddress())
+
+proc newWebsocketClient*(uri: Uri, protocols: seq[string] = @[]): Future[
+    WebSocket] {.async.} =
+  let
+    keyDec = align(
```
cheatfate, line 371: **SECURITY WARNING:** We are using epoch seconds as security key? Is this from WS specification, or just proper random number could be used?

src/http.nim
arijitAD: HTTP header is not case sensitive. Hence converting it to title case for better readability.

ws.nimble
arijitAD: Done

src/ws.nim
arijitAD: Fixed it to use local `./random`

src/ws.nim
arijitAD: Done

src/ws.nim
arijitAD: Done

src/ws.nim
arijitAD: Removed

arnetheduck:
> or any standard guidelines that we follow?

we've since written https://status-im.github.io/nim-style-guide/

src/ws.nim
arijitAD: Done

src/ws.nim
arijitAD: Done

src/ws.nim
arijitAD: Done

src/ws.nim
arijitAD: Added `MsgReader` that fills the buffer provided by the user of the library.
```
proc readMessage*(msgReader: MsgReader,data: seq[byte]): MsgReader {.async.} =
```

src/ws.nim
arijitAD: Done

src/http.nim
```diff
@@ -0,0 +1,250 @@
...
+  except TransportOsError as exc:
+    debug "Problems with networking", address = transp.remoteAddress(),
+          error = exc.msg
+    error = true
+
+  if error:
+    buffer.setLen(0)
+  return buffer
+
+proc newConnection(client: HttpClient, url: Uri) {.async.} =
```
dryajov, line 62: By convention, `newXXX` is used in place of constructors (tho, this is also discouraged by the most recent style guide). For the sake of avoiding confusion, I would just call this method `connect()`.

src/http.nim
```diff
@@ -0,0 +1,250 @@
...
+# Send request to the client. Currently only supports HTTP get method.
+proc request*(client: HttpClient, url, httpMethod: string,
+             body = "", headers: HttpHeaders): Future[seq[byte]]
+             {.async.} =
+  # Helper that actually makes the request. Does not handle redirects.
+  let requestUrl = parseUri(url)
+  if requestUrl.scheme == "":
+    raise newException(ValueError, "No uri scheme supplied.")
+
+  await newConnection(client, requestUrl)
```
dryajov, line 102:
```suggestion
  await client.connect(requestUrl)
```

For reference, Nim supports UFFC - https://en.wikipedia.org/wiki/Uniform_Function_Call_Syntax

src/http.nim
dryajov: Yeah, I meant that `toTitleCase` is replaced with `toUpper` or `toLower` since case is irrelevant here. `toTitleCase` seems to be doing too much by allocating a buffer and copying data just for the sake of capitalizing the first letter?

src/ws.nim
```diff
@@ -0,0 +1,451 @@
...
+    return
+  ws.readyState = Closed
+  await ws.send(@[], Close)
+  if initiator == true:
+    let frame = await ws.readFrame()
+    if frame.opcode != Close:
+      echo "Different packet type"
+  await ws.close()
+
+proc readMessage*(msgReader: MsgReader,data: seq[byte]): MsgReader {.async.} =
```
dryajov, line 350: Nim uses copy semantics, so this will create a copy of the buffer. You are better off by using a `pointer` and a `size` argument in this case, you can see how it's done in the libp2p code https://github.com/status-im/nim-libp2p/blob/master/libp2p/stream/lpstream.nim#L146-L149

In non-async code, it's possible to also use `var openArray[T]` arguments, which are preferable and will pass a "managed" reference of an object to the calling proc. `openArrays` is a generic type that can represent `seq` and `array`, you can read more about it here - https://nim-lang.org/docs/manual.html#types-open-arrays.

src/ws.nim
```diff
@@ -0,0 +1,451 @@
...
+       rsv2: false,
+       rsv3: false,
+       opcode: opcode,
+       mask: ws.masked,
+       data: data,
+       maskKey: maskKey)
+    var frame = encodeFrame(inFrame)
+    const maxSize = 1024*1024
+    # Send stuff in 1 megabyte chunks to prevent IOErrors.
+    # This really large packets.
```
dryajov, line 227: Why do we need to send this much data everytime, seems a bit too large?

src/ws.nim
```diff
@@ -0,0 +1,451 @@
...
+    # Read the mask.
+    await ws.tcpSocket.readExactly(addr maskKey[0], 4)
+    for i in 0..<maskKey.len:
+      frame.maskKey[i] = cast[char](maskKey[i])
+
+  if (frame.opcode == Text) or (frame.opcode == Opcode.Cont) or (frame.opcode == Opcode.Binary) :
+    return frame
+
+  # TODO: Add check for max size for control frames.
+  var data = newSeq[byte](finalLen)
```
dryajov, line 319: We need to add a max size for `finalLen`, otherwise an attacker can force us to allocate a very large buffer.

src/ws.nim
```diff
@@ -0,0 +1,451 @@
...
+    raise newException(WebSocketError, "Socket mask mismatch")
+
+  var maskKey = newSeq[byte](4)
+  if frame.mask:
+    # Read the mask.
+    await ws.tcpSocket.readExactly(addr maskKey[0], 4)
+    for i in 0..<maskKey.len:
+      frame.maskKey[i] = cast[char](maskKey[i])
+
+  if (frame.opcode == Text) or (frame.opcode == Opcode.Cont) or (frame.opcode == Opcode.Binary) :
```
dryajov, line 315: This should be commented to indicate that we only continue if the frame is `control` frame.

dryajov (approved):
I'll merge this as is and we can work on the remaining items in separate PRs.

dryajov:
Closing to address issues in separate PRs

PR #5, opened 2021-03-11
Rework

dryajov:
This is a minimal functional implementation that reworks:

- API, it adds two methods to receive data
  - `proc recv*(ws: WebSocket, data: pointer, size: int): Future[int] {.async.}`, allows passing a pointer to up to `size` bytes, note that this will potentially consume multiple and/or partial frames. This will consume a frame as soon as it arrives, without buffering, which allows streaming.
  - `proc recv*(ws: WebSocket, size = WSMaxMessageSize): Future[seq[byte]] {.async.}` a convenience method that accumulates frames up to a max message size
  - Note that in WS messages split into arbitrary frames delimited by a `fin` flag
- Fixes an important issue with framing, where a message would be stuffed into one frame, encoded and then split into chunks of 1MB, this is dangerous as anything else writing to the same pipe in between chunks will get interleaved with the frame chunks (this issue is also present in the nim-news package!), the correct behaviour is to split the data into smaller chunks and then frame that
- Fixed control message handling
- Fixed closing flow handling (NOTE, Websockets do not support half-closed/open flow)
- General cleanup and testing

This PR does not address any of the HTTP handling functionality already present, which is largely based on the `nim-json-rpc` implementation but should potentially be reworked with the new HTTP functionality recently added to chronos.


src/ws.nim
```diff
@@ -205,226 +280,361 @@ proc encodeFrame(f: Frame): seq[byte] =
...
+    ws.readyState = ReadyState.Closed
+  else:
+    raiseAssert("Invalid state during close!")
 
-  var opcode = b0 and 0x0f
-  if opcode notin WSOpCode:
-    raise newException(WebSocketError, "Unexpected  websocket opcode")
-  frame.opcode = (opcode).Opcode
+proc handleControl*(ws: WebSocket, frame: Frame) {.async.} =
+  ## handle control frames
```
arijitAD, line 376: We can add a check on the size of the control frame data. According to the [spec](https://tools.ietf.org/html/rfc6455#section-5.5.1):

> All control frames MUST have a payload length of 125 bytes or less

src/ws.nim
```diff
@@ -205,226 +280,361 @@ proc encodeFrame(f: Frame): seq[byte] =
...
-  var hf = cast[HeaderFlags](b0 shr 4)
-  frame.fin = fin in hf
-  frame.rsv1 = rsv1 in hf
-  frame.rsv2 = rsv2 in hf
-  frame.rsv3 = rsv3 in hf
+  let maxSize = ws.frameSize
+  # Send stuff in 1 megabyte chunks to prevent IOErrors.
+  # This really large packets.
+  var i = 0
+  while i < data.len:
```
arijitAD, line 317: We should return if the closing is initiated.

> The application MUST NOT send any more data frames after sending a Close frame.

src/ws.nim
```diff
@@ -205,226 +280,361 @@ proc encodeFrame(f: Frame): seq[byte] =
...
+        fin: if (i + len >= data.len): true else: false,
+        rsv1: false,
+        rsv2: false,
+        rsv3: false,
+        opcode: if i > 0: Opcode.Cont else: opcode, # fragments have to be `Continuation` frames
+        mask: ws.masked,
+        data: data[i ..< len],
+        maskKey: maskKey)
+
+    discard await ws.tcpSocket.write(encodeFrame(inFrame))
```
arijitAD, line 329: Shouldn't we verify the encoded frame length after writing to the socket or have an exception handler around the write?

src/ws.nim
```diff
@@ -205,226 +280,361 @@ proc encodeFrame(f: Frame): seq[byte] =
...
+      # return the current frame if it's not one of the control frames
+      if frame.opcode notin {Opcode.Text, Opcode.Cont, Opcode.Binary}:
+        asyncSpawn ws.handleControl(frame) # process control frames
+        continue
+
+      debug "Decoded new frame", opcode = frame.opcode, len = frame.length, mask = frame.mask
+
+      return frame
+  except CatchableError as exc:
+    debug "Exception reading frame, dropping socket", exc = exc.msg
```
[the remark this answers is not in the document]
dryajov: Well, since frames are ordered, we should not be reading any more frames after closing, so I think it makes sense to throw here either way.

src/ws.nim
dryajov: Ah, nice catch.

src/ws.nim
dryajov: We probably want the error from the write to propagate to the caller, we might want to handle it and close the socket/cleanup and then rethrow tho.

src/ws.nim
dryajov: Hmm, maybe we want to throw if writing to a closed/closing (ws) socket.

src/ws.nim
dryajov: For the lenght, as far as I can tell from the spec, it only cares about the payload?

sinkingsugar (approved):
Well that was a rework indeed.
LGTM, maybe adding coverage tracking would help as it definitely needs lot of more testing to find runtime issues.

src/ws.nim
```diff
@@ -205,226 +281,368 @@ proc encodeFrame(f: Frame): seq[byte] =
...
+  ## handle control frames
+  ##
 
-  var opcode = b0 and 0x0f
-  if opcode notin WSOpCode:
-    raise newException(WebSocketError, "Unexpected  websocket opcode")
-  frame.opcode = (opcode).Opcode
+  if frame.length > 125:
+    raise newException(WSPayloadTooLarge,
+      "Control message payload is freater than 125 bytes!")
```
arijitAD, line 383: typo: greater

arijitAD (approved):

dryajov:
This should be ready for review. As it is, it still lacks comprehensive tests, but it is a fully functional implementation with browser interop. 

The next steps are:

- HTTP rework and SSL support
- Adding more comprehensive tests
- Adding browser centric tests perhaps with karma or something similar 
  - the most important point is that browser interop is guaranteed

dryajov:
CI is broken for i386 for github actions here is an issue tracking it - https://github.com/actions/virtual-environments/issues/2919#issuecomment-799586341. The workaround didn't work, but it's not critical to merge.

PR #6, opened 2021-03-20
Update http to use chronos http.

arijitAD:


src/ws.nim
```diff
@@ -147,6 +150,19 @@ type
 template remainder*(frame: Frame): uint64 =
   frame.length - frame.consumed
 
+proc toString*(ht: HttpTables): string =
```
dryajov, line 153: Usually this is accomplished with the dollar (`$`) operator,  you can even leave the `toString` proc and just call it from the `$` proc.

```suggestion
proc toString*(ht: HttpTables): string =
...

proc `$`*(ht: HttpTables): string = ht.toString
```

When you want to convert to string you simply prepend the variable with the `$`, like so

```nim
echo $ht
```

src/stream.nim
```diff
@@ -0,0 +1,53 @@
...
+    error = true
+
+  if error:
+    buffer.setLen(0)
+  return buffer
+
+proc closeWait*(wsStream : AsyncStream): Future[void] {.async.} =
+  if not wsStream.writer.tsource.closed():
+    await wsStream.writer.tsource.closeWait()
+  if not wsStream.writer.tsource.closed():
```
dryajov, line 50:
```suggestion
  if not wsStream.reader.tsource.closed():
```

src/ws.nim
```diff
@@ -617,7 +630,42 @@ proc close*(
...
+  var transp: StreamTransport
+  var requestHeader: string
+  try:
+    transp = await connect(address)
+  except CatchableError as exc:
+    raise newException(TransportError, "Cannot connect to " & $transp.remoteAddress() & " Error: " & exc.msg)
+  
+  let reader = newAsyncStreamReader(transp)
+  let writer = newAsyncStreamWriter(transp)
+  requestHeader = "GET " & uri.path & " HTTP/1.1" & CRLF & headers.toString()
```
dryajov, line 648: Can be rewritten with the `$`

```suggestion
  requestHeader = "GET " & uri.path & " HTTP/1.1" & CRLF & $headers
```

src/ws.nim
arijitAD: Done

src/stream.nim
arijitAD: Done

src/ws.nim
arijitAD: Done

dryajov:
@arijitAD what's still left to be done in this PR, can you please add a TODO list?

Also, for convenience, we usually expose an `accept` call that can be invoked as `let conn = await ws.accept()`, this is not strictly necessary for this PR but it would be good to keep in mind because it will make using integration with other modules, including libp2p easier.

arijitAD:
> @arijitAD what's still left to be done in this PR, can you please add a TODO list?
> 
> Also, for convenience, we usually expose an `accept` call that can be invoked as `let conn = await ws.accept()`, this is not strictly necessary for this PR but it would be good to keep in mind because it will make using integration with other modules, including libp2p easier.

@dryajov The PR is ready for review. CI seems to be failing for windows. 
I have created an issue for libp2p intergration. https://github.com/status-im/nim-ws/issues/8

src/stream.nim
```diff
@@ -0,0 +1,53 @@
...
+    let ores = await withTimeout(hlenfut, HttpHeadersTimeout)
+    if not ores:
+      # Timeout
+      debug "Timeout expired while receiving headers",
+            address = rstream.tsource.remoteAddress()
+      error = true
+    else:
+      let hlen = hlenfut.read()
+      buffer.setLen(hlen)
+  except TransportLimitError:
```
cheatfate, line 29: `AsyncStreamReader` interface raising different exception types.
https://github.com/status-im/nim-chronos/blob/master/chronos/streams/asyncstream.nim#L27-L35

src/stream.nim
```diff
@@ -0,0 +1,53 @@
...
+    debug "Problems with networking", address = rstream.tsource.remoteAddress(),
+          error = exc.msg
+    error = true
+
+  if error:
+    buffer.setLen(0)
+  return buffer
+
+proc closeWait*(wsStream : AsyncStream): Future[void] {.async.} =
+  if not wsStream.writer.tsource.closed():
```
cheatfate, line 48: This is a little bit wrong. You assume here that `AsyncStream` is wrapper around `StreamTransport`, but its not always true.
When you get `AsyncStreamReader`, `AsyncStreamWriter` from HttpServer interface you can get wrappers only if connection is non-secure (plain HTTP), but if connection is secure `AsyncStreamReader/Writer` would be actually `TLSStreamReader/Writer` and corresponding `tsource` will be `nil`, while corresponding `rsource` will be `AsyncStreamReader/Writer` wrappers.

src/ws.nim
```diff
@@ -164,45 +180,42 @@ proc prepareCloseBody(code: Status, reason: string): seq[byte] =
...
   ws: WebSocket,
-  header: HttpRequestHeader,
+  request: HttpRequestRef,
   version = WSDefaultVersion) {.async.} =
   ## Handles the websocket handshake.
   ##
-
-  discard parseSaturatedNatural(header["Sec-WebSocket-Version"], ws.version)
+  let reqHeaders = request.headers
+  discard parseSaturatedNatural(reqHeaders.getString("Sec-WebSocket-Version"), ws.version)
```
cheatfate, line 188: Please use `stew/base10` for parsing integers. `parseSaturatedNatural` allow `_`, ` ` and `-`.

src/ws.nim
```diff
@@ -164,45 +180,42 @@ proc prepareCloseBody(code: Status, reason: string): seq[byte] =
...
   let cKey = ws.key & WSGuid
   let acceptKey = Base64Pad.encode(sha1.digest(cKey.toOpenArray(0, cKey.high)).data)
-
-  var response = "HTTP/1.1 101 Web Socket Protocol Handshake" & CRLF
-  response.add("Sec-WebSocket-Accept: " & acceptKey & CRLF)
-  response.add("Connection: Upgrade" & CRLF)
-  response.add("Upgrade: webSocket" & CRLF)
-
+  
+  var headerData = @[("Connection", "Upgrade"),("Upgrade", "webSocket" ),
```
cheatfate, line 205: There no need to create sequence with `@`, you can pass an array.

src/ws.nim
```diff
@@ -347,7 +361,7 @@ proc handleClose*(ws: WebSocket, frame: Frame) {.async.} =
     var data = newSeq[byte](frame.length)
     if frame.length > 0:
       # Read the data.
-      await ws.tcpSocket.readExactly(addr data[0], int frame.length)
+      await ws.stream.reader.readExactly(addr data[0], int frame.length)
```
cheatfate, line 364: Missing error handler?

src/ws.nim
```diff
@@ -363,13 +377,13 @@ proc handleClose*(ws: WebSocket, frame: Frame) {.async.} =
       try:
         (rcode, reason) = ws.onClose(code, string.fromBytes(data))
       except CatchableError as exc:
-        debug "Exception in Close callback, this is most likelly a bug", exc = exc.msg
+        debug "Exception in Close callback, this is most likely a bug", exc = exc.msg
 
-    # don't respong to a terminated connection
+    # don't respond to a terminated connection
     if ws.readyState != ReadyState.Closing:
       await ws.send(prepareCloseBody(rcode, reason), Opcode.Close)
```
cheatfate, line 384: Missing error handler?

examples/client.nim
```diff
@@ -1,8 +1,9 @@
-import ../src/ws, nativesockets, chronos, os, chronicles, stew/byteutils
+import ../src/ws, nativesockets, chronos,chronicles, stew/byteutils
 
 proc main() {.async.} =
-  let ws = await connect(
-    "127.0.0.1", Port(8888),
+  let ws = await wsConnect(
+    "127.0.0.1", 
+    Port(8888),
```
dryajov, line 6: I think defining `connect` as `proc connect(_: type WebSocket, ...)` is more nim'ish and it allows you to use it as a static proc - `WebSocket.connect(...)`

src/stream.nim
dryajov: @arijitAD why not just do `wsStream.closeWait()`?

dryajov: Or better yet, just leave this method out since the strems already expose this method and there is no reason for another level of indirection.

dryajov: @cheatfate, why doesn't chronos expose a sane `closeWait()` version to close this streams? It's a bit akward to have to care about all the subtleties of what is set when and on which version of the stream. 

Can't we just have `closeWait()` method that works for all this versions of the streams? I tried figuring out what the correct procedure to close them is, but it's deinitely not straightforward.

PR #7, opened 2021-03-23
WIP: Implement websocket TLS.

arijitAD:


examples/tlsclient.nim
```diff
@@ -0,0 +1,35 @@
...
+
+import ../src/ws
+ 
+proc main() {.async.} =
+    let ws = await webSocketTLSConnect(
+        "127.0.0.1",
+        Port(8888),
+        path = "/wss",
+        protocols = @["myfancyprotocol"],
+        flags = {NoVerifyHost,NoVerifyServerName})
```
cheatfate, line 14: This flags must not be defaults, because its insecure.

cheatfate:
I will recommend to add leaks testing at the end of test suite. 
https://github.com/status-im/nim-chronos/blob/master/tests/testasyncstream.nim#L1010-L1015
Because i suspect there present leaks.

cheatfate:
I will recommend to add leaks testing at the end of test suite. 
https://github.com/status-im/nim-chronos/blob/master/tests/testasyncstream.nim#L1010-L1015
Because i suspect there present leaks.

dryajov:
Yes, great idea. We need to add chronos trackers to ensure everything is
being released property.

examples/tlsclient.nim
dryajov: That's just an example implementation, it should be fine here.

cheatfate: Exactly you should not make examples with such defaults, because people will use your examples in their code.

tests/testtlswebsockets.nim
```diff
@@ -149,7 +149,6 @@ suite "Test websocket TLS transmission":
         check request.uri.path == "/wss"
         let ws = await createServer(request, "proto")
         await ws.send(testString)
-        await ws.stream.closeWait()
```
dryajov: Hmm, actually we should not call `closeWait()` on the stream - it should be private. We should only call `ws.close()`.

dryajov:
@arijitAD I did a quick pass to remove all the `stream.closeWait()` from tests, where it is relevant, but unfortunately it revealed some additional failure. Let's spend some more time on this and make sure all tests are sound and correct.

examples/tlsclient.nim
dryajov: Would this be possible in a test setup running against localhost/127.0.0.1? We need to make sure that this run locally, with as little setup as possible.

arijitAD:
@dryajov There were some issue with closing handshake which I fixed in this PR. https://github.com/status-im/nim-ws/pull/9
I removed all `stream.closeWait` and CI is passing.

dryajov:
> @dryajov There were some issue with closing handshake which I fixed in this PR. #9
> I removed all `stream.closeWait` and CI is passing.

Lets fix it here please before moving on to #9

dryajov (changes requested):
> @dryajov There were some issue with closing handshake which I fixed in this PR. #9
> I removed all `stream.closeWait` and CI is passing.

Lets fix it here please before moving on to #9

tests/testtlswebsockets.nim
```diff
@@ -124,9 +124,7 @@ suite "Test websocket TLS transmission":
         check request.uri.path == "/wss"
         let ws = await createServer(request, "proto")
         let servRes = await ws.recv()
-
         check string.fromBytes(servRes) == testString
-        await ws.close()
```
dryajov: This isn't the proper fix, removing this close just hides the real issue, which seems to be related to the tls server in chronos, running it with the regular http server works as expected. 

@cheatfate can you please take a look, the closeWait hangs on windows on the tls server after both sockets are closed, roving the closing of one of the sockets seems to remove the hang. I traces it down to where all the processLoop features are cancelled and awaited, the hangs happens right there.

arijitAD: Yes. This was a temporary hack. I assumed the issue was with WebSocket close. https://github.com/status-im/nim-ws/pull/9 Ported a couple of changes from this PR but the close didn't work. It could be an issue in the TLS server as you mentioned.

dryajov: @arijitAD, can you please create a minimal repro of this issue and _iff_ it still present, log it as a bug in chronos?

arijitAD: Sure. I will do that.

dryajov: Turns out that the server callback needs to return a `dumbRespose()` otherwise it will attempt to write a response on an already closed connection. This is likely the cause of the hang on windows, but it's pretty terrible UX in general. We should add an accept call to chronos or to this implementation to avoid this types of issues.

dryajov (approved):

PR #35, opened 2021-05-26
implement UTF8 handling

### LINKED ISSUE #13: Implement UTF-8 Handling

jangko:
or to be precisely, we add UTF8 validator which main duty
is to detect malformed UTF8 sequence using a fast DFA UTF8
decoder.

also enable autobahn UTF8 category tests, much more green :)

fixes #13

### DIFF
--- a/tests/testutf8.nim
+++ b/tests/testutf8.nim
@@ -1,0 +1,229 @@
+## nim-ws
+## Copyright (c) 2021 Status Research & Development GmbH
+## Licensed under either of
+##  * Apache License, version 2.0, ([LICENSE-APACHE](LICENSE-APACHE))
+##  * MIT license ([LICENSE-MIT](LICENSE-MIT))
+## at your option.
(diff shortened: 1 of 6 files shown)


dryajov (commented):
Can we please add a couple of unit tests? Otherwise, LGTM 👍

tests/testutf8.nim
```diff
@@ -0,0 +1,235 @@
...
+        path = "/ws",
+        protocols = @["proto"]
+      )
+
+      await wsClient.send(testData)
+      await wsClient.close(reason = closeReason)
+      await waitForClose(wsClient)
+      check wsClient.readyState == ReadyState.Closed
+
+suite1()
```
dryajov, line 233: The `suite` and `test` blocks should be automatically invoked, any reason for the explicit invocation?

jangko: only my old habit. unittest used to pollute global space with too much variables and causing strange behavior.
I guess asynctest is different.

PR #48, opened 2021-06-01
wip: modeling extensions api

dryajov:
Modeling the extensions API - @jangko let me know what you think.

### DIFF
--- a/ws/frame.nim
+++ b/ws/frame.nim
@@ -55,7 +55,7 @@ template remainder*(frame: Frame): uint64 =
 proc encode*(
   frame: Frame,
   offset = 0,
-  extensions: seq[Extension] = @[]): Future[seq[byte]] {.async.} =
+  extensions: seq[Ext] = @[]): Future[seq[byte]] {.async.} =
   ## Encodes a frame into a string buffer.
   ## See https://tools.ietf.org/html/rfc6455#section-5.2
 
@@ -117,7 +117,7 @@ proc decode*(
   _: typedesc[Frame],
   reader: AsyncStreamReader,
   masked: bool,
-  extensions: seq[Extension] = @[]): Future[Frame] {.async.} =
+  extensions: seq[Ext] = @[]): Future[Frame] {.async.} =
   ## Read and Decode incoming header
   ##
 
--- a/ws/session.nim
+++ b/ws/session.nim
@@ -61,7 +61,7 @@ proc send*(
         mask: ws.masked,
         data: data, # allow sending data with close messages
         maskKey: maskKey)
-        .encode()))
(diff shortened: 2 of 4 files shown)


jangko:
the api looks good, it's very similar to my compression branch.
one thing seems not right is about the extension execution chain. but that can be fixed later, because it need quite extensive testing involving more than one extension, although not necessarily a compression extension. 
we can mix compression and some simple text transformation e.g uppercase or lower case extension.

see https://datatracker.ietf.org/doc/html/rfc7692.html#section-6.1

dryajov:
> one thing I'm not so sure is about the extension execution chain. but that can be fixed later, 

Yeah, thats the next thing I want to wrap up here. The spec is a bit convoluted in the way it states the extension selection procedure, here the relevant part:

(https://datatracker.ietf.org/doc/html/rfc6455#section-9)

```
   Note that the order of extensions is significant.  Any interactions
   between multiple extensions MAY be defined in the documents defining
   the extensions.  In the absence of such definitions, the
   interpretation is that the header fields listed by the client in its
   request represent a preference of the header fields it wishes to use,
   with the first options listed being most preferable.  The extensions
   listed by the server in response represent the extensions actually in
   use for the connection.  Should the extensions modify the data and/or
   framing, the order of operations on the data should be assumed to be
   the same as the order in which the extensions are listed in the
   server's response in the opening handshake.

   For example, if there are two extensions "foo" and "bar" and if the
   header field |Sec-WebSocket-Extensions| sent by the server has the
   value "foo, bar", then operations on the data will be made as
   bar(foo(data)), be those changes to the data itself (such as
   compression) or changes to the framing that may "stack".
```

Specifically this part seems to allude to the existence of some notion of priorities/order:

```
Any interactions between multiple extensions MAY be defined in the documents defining the extensions.
```

For example if both extensions `A` and `B` are registered, if `A` > `B`, then the server always responds with `A` and then `B`, regardless of the order in which the client requested this extensions

One way to implement this would then be:

- If no "priority" is defined, the server should select the extensions in the order in which the client requested them
- If a "priority" is defined, the server will respond with the extensions that it supports using a combination of both, the priority of the extension if it defines one and the order in which the client requested the extensions

> need quite extensive testing involving more than one extension, although not necessarily a compression extension.

Yeah, testing can be done with a dummy extension and the compression one

The one thing I'm not 100% clear on yet is if extensions can read from the stream, or they only process whatever data is already in the frame? Maybe this is what you're referring to here with:

`it means there is possibility we need to change encode and decode api. or probably a flag for extension instance is enough.` ?

In any case it should be easy to support, we simply pass the `AsyncStreamReader/Writer` to the extensions.

Issue #32, opened 2021-05-26
|Host| field is required in handshake


jangko:
[rfc6455#section-4.2.1](https://datatracker.ietf.org/doc/html/rfc6455#section-4.2.1) point 2.

> 2.   A |Host| header field containing the server's authority.

in nim-ws, both server and client seems ignoring this field.

error message I get when try to run `autobahn-client.nim` vs autobahn-fuzzingserver:
```text
DBG 2021-05-26 10:21:27.493+07:00 Websocket failed during handshake          tid=11672 exc="Server did not reply with a websocket upgrade: Header code: 400 Header reason: HTTP Host header missing in opening handshake request Address: 127.0.0.1:9001"
ERR 2021-05-26 10:21:27.495+07:00 WebSocket error                            tid=11672 exception="Server did not reply with a websocket upgrade: Header code: 400 Header reason: HTTP Host header missing in opening handshake request Address: 127.0.0.1:9001"
```

PR #49, opened 2021-06-02
Implement compression

jangko:


ws/ws.nim
```diff
@@ -260,7 +304,7 @@ proc handleRequest*(
 proc new*(
   _: typedesc[WSServer],
   protos: openArray[string] = [""],
-  factories: openArray[ExtFactory] = [],
+  factories: openArray[(string, ExtFactory)] = [],
```
dryajov, line 307: Was thinking about this. Maybe just pass a `Table` instead (not sure)?

dryajov:
`ExtParam` should probably be moved to `types.nim`.

ws/ws.nim
```diff
@@ -183,6 +183,45 @@ proc connect*(
     onClose = onClose,
     rng = rng)
 
+proc selectExt(ws: WSServer, request: HttpRequest): string
+  {.raises: [Defect, WSExtError].} =
+
+  # it is possible to have multiple "Sec-WebSocket-Extensions"
+  let exts = request.headers.getList("Sec-WebSocket-Extensions")
+  var extList: seq[AppExt]
```
dryajov, line 191: I think `AppExt` is the same as `Ext`, should we merge them?

ws/types.nim
```diff
@@ -92,10 +92,9 @@ type
 
   Ext* = ref object of RootObj
     name*: string
-    options*: Table[string, string]
 
-  ExtFactory* = proc(name: string, options: Table[string, string]):
-    Ext {.raises: [Defect].}
+  ExtFactory* = proc(serverSide: bool, options: seq[ExtParam]):
```
dryajov, line 96: Do we really need to make a distinction for server vs client side?

ws/ws.nim
jangko: I have strong reason to keep the `openArray` interface.

1. not sacrifice ergonomics. I imagine authors of extensions would export their extensions using:
```Nim
const
    deflateExt* = ("permessage-deflate", deflateFactory)

const
    snappyExt* = ("permessage-snappy", snappyFactory)

etc..

let ws = WSServer.new(factories = [deflateExt, snappyExt])  # still ergonomic
```

2. flexibility to swap internal implementation
I assume in real life production code, the number of extensions is not very much. probably not much than 5.
If that is the case, it is more economically to use plain `seq` instead of `Table`. Table is overkill if number of exts is small.
keeping the `openArray` interface will allow us to change the implementation without breaking the interface.

ws/ws.nim
jangko: I agree ExtParam should be moved to `types.nim`.

My idea to separate `AppExt` and `Ext` is to separate server code(ext creation) and parser code. If we want to merge them, then the parser code and server code will be interleaving.  But I will revisit this later to find a clean solution.

ws/types.nim
jangko: currently I'm using it for optimization and diagnosis purpose, I need more thorough analysis before discard it.

ws/ws.nim
dryajov: > I have strong reason to keep the openArray interface.

OK, I like the reasoning and I don't really have a strong objection against this.

> If that is the case, it is more economically to use plain seq instead of Table. Table is overkill if number of exts is small.

Yeah agree with that, I believe I had it as a `seq` initially?

ws/ws.nim
dryajov: I think that for the most part, the client/server functionality is very similar, if we can merge them I'd prefer we do that.

ws/ws.nim
jangko: yes, and I will change it back to seq.

jangko:
@dryajov , the server side compression is almost  ready. maybe you can review it again while I prepare the miniz repo.
for the client side compression, that will be in the next PR.

ws/ws.nim
```diff
@@ -183,6 +183,64 @@ proc connect*(
     onClose = onClose,
     rng = rng)
 
+proc extInList(ws: WSServer, extName: string): bool =
```
dryajov, line 186: `contains` would be more idiomatic here

ws/compression.nim
```diff
@@ -0,0 +1,212 @@
...
+  for i in 0..<arg.value.len:
+    winbit = winbit * 10 + arg.value[i].int - '0'.int
+
+  if winbit < 8 or winbit > 15:
+    return err("window bits should between 8-15, got " & $winbit)
+
+  res = winbit
+  return ok("=" & arg.value)
+
+proc validateParams(args: seq[ExtParam],
```
dryajov, line 60: Would something like `createParams` reflect it's intent better?

dryajov:
> @dryajov , the server side compression is almost ready. maybe you can review it again while I prepare the miniz repo.
> for the client side compression, that will be in the next PR.

@jangko I spent some time looking into how extensions and specifically per message deflate works and this PR needs to be reworked to some extent. 

The core of the issue is that extensions can/should hijack the flow, as it is the case of the per message deflate as outlined here - https://datatracker.ietf.org/doc/html/rfc7692#section-6:

- Compressing implies that the message is compressed and then passed to the next extension in the chain, per message deflate is already the last extension, then it will frame the compressed message.
- Decompressing implies buffering the message, decompressing it and then passing the body and/or frame to the next extension in the chain or, the application, if per message deflate is the last extension.

The way it is implemented now, won't work 100% of the cases. For example calling `decode(...)` [here](https://github.com/status-im/nim-ws/pull/49/files#diff-06f7698d7319a084ba0ec1981ccdaa2f650026e437b179ab5afe22e5f67745b9R352) is the wrong thing to do as this is only a convenience proc to avoid having to deal with pointers and such.

In short we need to rework some of the flow to accommodate for this requirements. I don't think this would affect the core functionality of this PR and we can rework the websockets flow in isolation. Once that's done, we can rebase and merge this changes.

dryajov (changes requested):
I'll leave it blocked for now to get the required changes to the extensions flow before merging this one in.

jangko:
actually my current approach is not working at all. 
the spec not only require specific flow in websocket implementation, but also specific way *to use* the compression library which is quite confusing initially -- and take significant amount of time to understand what is going on.

1. after some experiments, it seems your original approach to the extension is already correct. -- compression done at per frame basis, compression flushing is important.
2. the websocket compression is designed around zlib interface and behavior. we can only use compression lib which support zlib behavior. in this case we still can use miniz. 
3. something like https://github.com/status-im/nim-zippy is unusable without serious modification.
    - it doesn't support LZ77 window size adjustment
    - it doesn't support segmented buffer for compression and decompression. all must done in one go.

dryajov:
> after some experiments, it seems your original approach to the extension is already correct. -- compression done at per frame basis, compression flushing is important. and compressor lib must support segmented buffer.

Yes, this is what I noticed as well, fortunately at least at the websockets level it doesn't seem to hard to modify `WSSession` and `Frame` to do the right thing.

>  we can only use compression lib which support zlib behavior. in this case we still can use miniz.

OK, good to know we can still use miniz - is it already wrapped? If not would zlib be more appropriate?

https://github.com/status-im/nim-zippy seems to be a pure Nim implementation - how hard would it be to extend it to support lz77 windowing?

jangko:
we already have https://github.com/status-im/nim-miniz.

> how hard would it be to extend it to support lz77 windowing?

easy.

but the segmented buffer and flushing mechanism is hard, because we need to redesign many of it's internal flow from contiguous operation into `pause-resume` operation.

ws/ws.nim
jangko: fixed

ws/compression.nim
jangko: fixed

jangko:
the server side compression now works.
it pass all autobahn tests section 12 and 13 cleanly.
no test cases because i still need to implement client compression in next PR.
we still cannot test client using autobahn due to #32.

dryajov:
> the server side compression now works.
> it pass all autobahn tests section 12 and 13 cleanly.
> no test cases because i still need to implement client compression in next PR.
> we still cannot test client using autobahn due to #32.

I believe `Host` is optional, why is it blocking the tests?

jangko:
> I believe Host is optional, why is it blocking the tests?

Fixes in this branch https://github.com/status-im/nim-ws/tree/auto_client allow autobahn_client to run. It sits on top of your PR(#56).

Don't know why autobahn server need 'Host' field. perhaps a bug.

ws/ws.nim
```diff
@@ -250,17 +313,23 @@ proc handleRequest*(
     raise newException(WSHandshakeError,
         "Failed to sent handshake response. Error: " & exc.msg)
 
-  return WSSession(
+  let session = WSSession(
     readyState: ReadyState.Open,
     stream: request.stream,
     proto: protocol,
+    extensions: system.move(ws.extensions),
```
dryajov, line 320: I think `extensions` is shared across all sessions in server mode?

dryajov: I guess I'm a bit confused about the `move` here, why do we need it?

jangko: no, it is not shared. the server will create new one for each session.

jangko: for extension such as `permessage-deflate`, as it's name implies. will keep a de/compression context from the start of message to the end of message. from first frame to last frame will use the same context. so extensions cannot be shared among session.

jangko: for a simple extension without configuration parameter, may the factory can keep a singleton to be shared among session.
but again, the extension chain itself can be different from one session to the next.

dryajov: > for extension such as permessage-deflate, as it's name implies. will keep a de/compression context from the start of message to the end of message. from first frame to last frame will use the same context. so extensions cannot be shared among session.

OK right, that makes sense then.

jangko:
current status of this PR:
- server side compression already pass all autobahn tests
- client side compression have many failing test cases because of:
   * autobahn server reject our extensions offer and force us to use unsupported compression feature(not supported by miniz).
   * succesfully negotiated compression configuration failing for mysterious reason. what I don't understand is the same test data    is used by both server and client mode, but the result/behavior is different.
  
will experiment more using zlib and see what is actually going on.

dryajov:
> not supported by miniz

Would it make sense to reconsider miniz in favor of a more complete implementation, or even just a native nim one? I much rather  have control over the entirety of the compression stack specially when already have a partial nim implementation.

jangko:
sure, we need to find a more complete deflate lib. but so far only zlib the strongest candidate. other nim deflate libs are not suitable or need too much modification.

jangko:
finally we have a working compression extension for both client and server.

PR #54, opened 2021-06-08
fixes multi frames message reading

### LINKED ISSUE #50: recv big binary message in multiple frames is buggy
sending a binary message with size = WSDefaultFrameSize * 3.

on recv side:
- the `binary` flag is not set, should be set.
- it will trigger `invalid UTF8 sequence detected`, should not happen.
- session terminated prematurely, should not happen.

identified bug:
https://github.com/status-im/nim-ws/blob/e632202037a6620d485d535f1a8ff2361cbaf5cb/ws/session.nim#L272

that line should not be there, it should be in the `if` branch after the first frame arrived.
because if next frame come, and it is a continuation frame, the binary flag will be unset.
the correct behavior is the binary flag is set for the entire session.

unidentified bug:
but after I move the line there, other bug appears. the data not completely received.
or to be precisely, the premature session termination is still there indicating another serious bug when dealing with continuation frames.

jangko:
instead of let the `frame.data` field empty
and read the payload to separate buffer,
now we read the payload into `frame.data` field.

this fixes also prepare for extensions and compression
that expect to manipulate a frame with it's payload.

fixes #50

### DIFF
--- a/ws/session.nim
+++ b/ws/session.nim
@@ -68,7 +68,7 @@ proc send*(
   let maxSize = ws.frameSize
   var i = 0
   while ws.readyState notin {ReadyState.Closing}:
-    let len = min(data.len, (maxSize + i))
+    let len = min(data.len - i, maxSize)
     await ws.stream.writer.write(
       (await Frame(
         fin: if (i + len >= data.len): true else: false,
@@ -77,7 +77,7 @@ proc send*(
         rsv3: false,
         opcode: if i > 0: Opcode.Cont else: opcode, # fragments have to be `Continuation` frames
         mask: ws.masked,
-        data: data[i ..< len],
+        data: data[i ..< i+len],
         maskKey: maskKey)
         .encode()))
 
@@ -258,6 +258,8 @@ proc recv*(
         if ws.frame.opcode == Opcode.Cont:
           raise newException(WSOpcodeMismatchError,
             "Expected Text or Binary frame")
+        
+        ws.binary = ws.frame.opcode == Opcode.Binary # set binary flag
       elif (not ws.frame.fin and ws.frame.remainder() <= 0):
         ws.frame = await ws.readFrame()
         # This could happen if the connection is closed.
@@ -268,8 +270,7 @@ proc recv*(
         if ws.frame.opcode != Opcode.Cont:
           raise newException(WSOpcodeMismatchError,
             "Expected Continuation frame")
-
-      ws.binary = ws.frame.opcode == Opcode.Binary # set binary flag
+      
       if ws.frame.fin and ws.frame.remainder() <= 0:
         ws.frame = nil
         break
--- a/tests/testwebsockets.nim
+++ b/tests/testwebsockets.nim
@@ -793,6 +793,38 @@ suite "Test Payload":
       ping
       pong
 
+  test "Send text message with multiple frames":
+    const FrameSize = 3000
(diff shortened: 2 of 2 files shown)


ws/session.nim
```diff
@@ -206,105 +206,70 @@ proc handleControl*(ws: WSSession, frame: Frame) {.async.} =
...
-    if frame.opcode notin {Opcode.Text, Opcode.Cont, Opcode.Binary}:
-      await ws.handleControl(frame) # process control frames# process control frames
-      continue
-
-    return frame
-
-proc ping*(ws: WSSession, data: seq[byte] = @[]): Future[void] =
-  ws.send(data, opcode = Opcode.Ping)
-
-proc recv*(
```
dryajov, line 229: @jangko we need this method so that we can pass an arbitrary buffer to read data into, otherwise this implies way to many allocations and it wont be easilly addaptable to libp2p.

ws/session.nim
```diff
@@ -206,105 +206,70 @@ proc handleControl*(ws: WSSession, frame: Frame) {.async.} =
...
-  ## to fill the buffer until either
-  ## the message ends (frame.fin) or
-  ## the buffer is full. If no data is on
-  ## the pipe will await until at least
-  ## one byte is available
-  ##
-
-  var consumed = 0
-  var pbuffer = cast[ptr UncheckedArray[byte]](data)
+  var buf: array[0xFFFF, byte]
```
dryajov, line 213: We can't assume that we only work at message boundaries, some applications might benefit from receiving frames as they come.

ws/session.nim
```diff
@@ -68,7 +68,7 @@ proc send*(
   let maxSize = ws.frameSize
   var i = 0
   while ws.readyState notin {ReadyState.Closing}:
-    let len = min(data.len, (maxSize + i))
+    let len = min(data.len - i, maxSize)
```
dryajov, line 71: This seems like the actual bug for multiple frames.

dryajov (changes requested):
I don't think we can use this as is because it makes a few assumptions that aren't necessarily correct for all cases. 

- It assumes that all extensions work at message boundaries, which isn't necessarily the case
- It prevents applications that can benefit from working at frame boundaries from doing so 
  - For example, when streaming non compressed data, it's possible to consume and process it as it comes, without waiting for the entire message to be accumulated
- Message and frames are also arbitrary websocket level concepts that might have little to do with how applications read/write data
  -  For example, an application might choose to send data in parts calling `send` several times for a large sequential chunk of data, which will then be split into frames at the websocket level, the receiving application can then attempt to read the entirety of the application level message by calling `recv(buf, fullDataSize)` which might span several message and frame boundaries, leaving that for the application to handle would imply leaking unrelated concepts such as frames and messages

The idea of passing around a frame to accumulate data into is valid, but we need to make sure we make as few assumptions about the structure and flow as possible

jangko:
ok, I get the big picture. will fix it.

ws/session.nim
jangko: restored

ws/session.nim
jangko: fixed

jangko:
this PR now only contains multi frames message fixes.
I will explore a new approach to extensions problem in other PR.

jangko:
closing in favor of #56

PR #56, opened 2021-06-09
Fix partial frame handling and allow extensions to hijack the flow

dryajov:
This PR addresses:

- [x] Partial frame reading which allows to stream data instead of buffering when possible
  - `proc recv*(ws: WSSession, data: pointer, size: int): Future[int] {.async.}` allows reading partial frames which allows streaming data more efficiently, this was always the intent but it wasn't fully implemented 
- [x] Handle close flow correctly
- [x] Adds support for extensions to hijack the frame processing flow
  - This ends up being really simple as we simply allow extensions to call `readFrame` with their own extensions list, including itself, which allows for flows imposed by extensions such as per-message compress, this should also be flexible enough to allow any possible flow
- [x] General cleanup
  - Rework status codes as distincts
  - Move extensions handling files under its own director
- [x] Autobahn should pass tests with regular http connections
- [ ] Autobahn should pass tests with tls http connections
  - Some tests are not closing connections cleanly
- [x] Test extensions flow using simple dummy extensions


dryajov:
@jangko I took the liberty of moving your test for broken multiple frames handling here, they are now passing. I believe the issue is now mitigated. I'll probably merge this in and work the remaining issues in a separate PR.

PR #59, opened 2021-06-12
fix limits tests

dryajov:


ws/session.nim
```diff
@@ -313,12 +313,19 @@ proc recv*(
   var pbuffer = cast[ptr UncheckedArray[byte]](data)
   try:
     var first = true
-
-    # reset previous frame if nothing is left in it
-    if not isNil(ws.frame) and ws.frame.remainder <= 0:
-      trace "Resetting previous frame"
-      first = ws.frame.fin # set as first frame if last frame was final
-      ws.frame = nil
+    if not isNil(ws.frame):
```
dryajov, line 316: can be simplified, but took some time to get right, so leaving here for now.

PR #60, opened 2021-06-12
fixes required to run autobahn client tests

jangko:


ws/ws.nim
```diff
@@ -61,7 +61,7 @@ proc connect*(
   ##
 
   var rng = if isNil(rng): newRng() else: rng
-  var key = Base64.encode(genWebSecKey(rng))
+  var key = Base64Pad.encode(genWebSecKey(rng))
```
dryajov, line 64: Does it have to be padded, I don't remember seeing this in the spec?

dryajov (approved):
LGTM

ws/ws.nim
jangko: it is not mentioned directly, but all `Sec-WebSocket-Key` always include padding char `=`. probably they assume developers already understand what is the meaning of this `=` char.

the autobahn server also reject our non-padded `Sec-WebSocket-Key` with message something like `..ASCII 24 chars...`.
it need fixed length `Sec-WebSocket-Key`

jangko:
btw, the CI shows regression in `handleControl`. 
it works fine when https://github.com/status-im/nim-ws/pull/56/commits/b44c049191730efa23e70e4d5d837d26bcdbb9b2

don't know whic one is the bad commit

ws/ws.nim
dryajov: Yeah, more like they assume base64 to be always padded on the network, not a bad assumption but without any practical implication wince headers are well delimited and there is no danger of collapsing base64 streams.

dryajov:
> btw, the CI shows regression in `handleControl`.
> it works fine when [b44c049](https://github.com/status-im/nim-ws/commit/b44c049191730efa23e70e4d5d837d26bcdbb9b2)
> 
> don't know whic one is the bad commit

OK, I'll take a look on monday.

dryajov:
> btw, the CI shows regression in `handleControl`.
> it works fine when [b44c049](https://github.com/status-im/nim-ws/commit/b44c049191730efa23e70e4d5d837d26bcdbb9b2)
> 
> don't know whic one is the bad commit

#62  partially addresses the issue, but something else is failing on windows.

PR #65, opened 2021-06-15
fixes related to extensions

jangko:
- set RSV bits in frame.encode
- move ExtParams from extutils.nim to types.nim
- remodel extension factory type
- accept/reject extensions offer in server
- offer/accept extensions in client

This PR is cherrypicked from compression experiments, from what we have learned from compression.

### DIFF
--- a/ws/types.nim
+++ b/ws/types.nim
@@ -9,8 +9,7 @@
 
 {.push raises: [Defect].}
 
-import std/tables
-import pkg/[chronos, chronos/streams/tlsstream]
+import pkg/[chronos, chronos/streams/tlsstream, stew/results]
 import ./utils
 
 const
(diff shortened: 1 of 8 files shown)


ws/types.nim
```diff
@@ -96,13 +95,21 @@ type
...
+
+  ExtFactoryProc* = proc(
+    isServer: bool,
+    args: seq[ExtParam]): Result[Ext, string] {.
+      gcsafe, raises: [Defect].}
+
+  ExtFactory* = object
+    name*: string
+    factory*: ExtFactoryProc
+    clientOffer*: string
```
dryajov, line 112: Just curious, why do we need this?

dryajov (approved):
LGTM

ws/types.nim
jangko: 1. `openArray[(string, ExtFactoryProc)]` does not work for closure proc, we need closure proc.
2. why we need closure proc? we need to capture extension configuration from user
3. keep it ergonomics
4. `clientOffer` is needed before the actual extension creation. the server also can rejects or accepts with configuration different from client offered.

```nim
proc extFactoryMaker(offerPadding: bool, useStdLib: bool): ExtFactory =
   proc factoryProc(.....): Ext =
        # capture extension configuration
        # e.g. `useStdLib`
        
   ExtFactory (
      name : "base64",
      factory: factoryProc,
      clientOffer: "base64; pad=" & $offerPadding  # client configuration
   )
```

dryajov: I was mostly asking about `clientOffer`, but thanks for the broader explanation.

PR #70, opened 2021-06-21
CI jobs parallelization to reduce autobahn test time

jangko:
with the inclusion of compression extension,
the autobahn test suite took significant time
to pass. move it to manually triggered CI
as it cannot make the main CI to fail anyway.

we only run basic autobahn tests in regular CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,12 @@
 name: CI
 on: [push, pull_request]
 
+# we dont run compression extension
+# in regular CI because it is a
+# time consuming operation.
+# we delegate it to manually triggered CI
+# and only run base autobahn tests here.
+
 jobs:
   build:
     strategy:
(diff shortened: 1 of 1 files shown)


jangko:
hmm, why the regular CI does not work when the manually triggered CI present?

dryajov:
How long does it take to run the tests? 

I prefer that we run all the tests all the time - because it allows us to catch issues quickier. The tradeoff is negligible +/-5 mins of CI time vs quickly detecting regressions - I'll take the later every time.

jangko:
> How long does it take to run the tests?

more than two hours https://github.com/status-im/nim-ws/actions/runs/954027726

jangko:
looks like github action computer already run too much parallel jobs. if I run the same test on my local computer with similar setup, the test will finish in less than an hour. but in github action it took more than two hours.

jangko:
New approach to make the CI run faster:
- 1st job: execute regular tests
- 2nd job: execute autobahn test in parallel with 1st job. also contains matrix for 4 test configuration(ws server, ws client, wss server, wss client).
- 3rd job: publish autobahn result after 2nd job completed using upload-download artifact.

dryajov:
Any progress on this @jangko ?

jangko:
I will do it today.

jangko:
hm, the tls client failed with error, but the CI pass, looks like we still have bug.

```bash
Error: unhandled exception: Unable to connect to host on any address! [HttpError]
```

dryajov:
> hm, the tls client failed with error, but the CI pass, looks like we still have bug.
> 
> ```shell
> Error: unhandled exception: Unable to connect to host on any address! [HttpError]
> ```

Interesting, I can't repro locally with latest `main`, was this a fluke?

jangko:
me too, can not reproduce it locally. but i've seen it twice in CI.

jangko:
get it, it's a bug in CI script. when I copy paste the previous code, I forget to remove the `&` from tls_client execution code. but it is still interesting because the same code did not fail in previous configuration.
why it has to fail when executing tls_client in background. need to wait for the client? probably.

dryajov:
> why it has to fail when executing tls_client in background. need to wait for the client? probably.

If there is no test server running, it will just fail connecting and drop out, no surprises there.

PR #71, opened 2021-06-23
update .nimble [skip ci]

jangko:
- change nim dependency from "1.2.6" to "1.2.0" because
  nimble refuse to install it when we are using
  nimbus-build-system
- fixes skipdir entry

### DIFF
--- a/ws.nimble
+++ b/ws.nimble
@@ -8,13 +8,13 @@
 ## those terms.
 
 packageName = "ws"
-version = "0.1.0"
-author = "Status Research & Development GmbH"
+version     = "0.1.0"
+author      = "Status Research & Development GmbH"
 description = "WS protocol implementation"
-license = "MIT"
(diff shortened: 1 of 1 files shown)


ws.nimble
```diff
@@ -8,13 +8,13 @@
...
+version     = "0.1.0"
+author      = "Status Research & Development GmbH"
 description = "WS protocol implementation"
-license = "MIT"
-skipDirs = @["examples", "test"]
+license     = "MIT"
+skipDirs    = @["examples", "tests"]
 
-requires "nim >= 1.2.6"
+requires "nim >= 1.2.0" # nimble will fail to install nim-ws if we are using 1.2.6 here
```
dryajov, line 17: Why is this?

jangko: it is a nimble bug that has been around for long time and still unsolved.

PR #75, opened 2021-06-25
fix compile time error when using chronicles sink:json

### LINKED ISSUE #73: Doesn't work with json logs
https://github.com/status-im/nim-libp2p/pull/593/checks?check_run_id=2905069466

Will also report it in chronicles, but it would probably be faster to fix it here (eg, not log addresses, or `$` them before-hand)

jangko:
fixes #73

### DIFF
--- a/ws/http/server.nim
+++ b/ws/http/server.nim
@@ -15,6 +15,10 @@ import pkg/[
   chronicles,
   httputils]
 
+#required to log addresses in chronicle's json mode
+import json_serialization/std/net
+
+
 import ./common
 
 type
@@ -38,13 +42,13 @@ proc validateRequest(
   ##
 
   if header.meth notin {MethodGet}:
-    trace "GET method is only allowed", address = $stream.tsource.remoteAddress()
+    trace "GET method is only allowed", address = stream.tsource.remoteAddress()
     await stream.sendError(Http405, version = header.version)
     return ReqStatus.Error
 
   var hlen = header.contentLength()
   if hlen < 0 or hlen > MaxHttpRequestSize:
-    trace "Invalid header length", address = $stream.tsource.remoteAddress()
+    trace "Invalid header length", address = stream.tsource.remoteAddress()
     await stream.sendError(Http413, version = header.version)
     return ReqStatus.Error
 
--- a/ws/http/server.nim
+++ b/ws/http/server.nim
@@ -38,13 +38,13 @@ proc validateRequest(
   ##
 
   if header.meth notin {MethodGet}:
-    trace "GET method is only allowed", address = stream.tsource.remoteAddress()
+    trace "GET method is only allowed", address = $stream.tsource.remoteAddress()
     await stream.sendError(Http405, version = header.version)
     return ReqStatus.Error
 
   var hlen = header.contentLength()
   if hlen < 0 or hlen > MaxHttpRequestSize:
-    trace "Invalid header length", address = stream.tsource.remoteAddress()
+    trace "Invalid header length", address = $stream.tsource.remoteAddress()
     await stream.sendError(Http413, version = header.version)
     return ReqStatus.Error
 
--- a/ws/http/server.nim
+++ b/ws/http/server.nim
@@ -15,8 +15,8 @@ import pkg/[
   chronicles,
   httputils]
 
-#required to log addresses in chronicle's json mode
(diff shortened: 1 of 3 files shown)


arnetheduck:
not that it matters much in trace logs, but calling `$`  explicitly is _slightly_ less efficient than letting chronicles deal with the formatting, thus we _generally_ avoid it (it allocates an extra string that chronicles in the future will be able to avoid for many types)

Menduist:
Nice, thanks for the quick fix :+1: we can revert this later when https://github.com/status-im/nim-chronicles/issues/98 is fixed, and maybe use ord instead of $ for the status code to avoid the extra string alloc in the future? But afaik this can be merged

Menduist:
Updated following conversation in https://github.com/status-im/nim-chronicles/issues/98

ws/http/server.nim
```diff
@@ -15,6 +15,10 @@ import pkg/[
   chronicles,
   httputils]
 
+#required to log addresses in chronicle's json mode
+import json_serialization/std/net
```
jangko, line 19: shouldn't we import json_serialization if only chronicles sink to json is defined?

Menduist: Yes, though not sure how to do this cleanly, eg
```nim
when config.streams[0].sinks[0].format == json:
  import json_serialization/std/net
```
This is horrible but works, we would need a helper function to check all sinks of all streams, and there doesn't seem to be one present in chronicles

jangko: I think you only need to check for this `chronicles_sinks == "json"`, it is exported from chronicles.

https://github.com/status-im/nim-chronicles/blob/63ce43a86a40a4c73d1b3b8317278d47ec55a458/chronicles/options.nim#L14

Menduist: Oh, I thought they were other ways to create streams that would have a json sink, but sure i'll do that

arnetheduck: there can be multiple sinks however

arnetheduck: and there can be `chronicles_streams` that write json

Menduist: https://github.com/status-im/nim-chronicles/pull/99
When this is merge, we can just do
```nim
when config.isLogFormatUsed(json):
  import json_serialization/std/net```

File: ws/session.nim [hunk 2]
```diff
@@ -184,7 +184,7 @@ proc handleClose*(
       raise newException(WSInvalidUTF8,
         "Invalid UTF8 sequence detected in close reason")
 
-  trace "Handling close message", code, reason
+  trace "Handling close message", code = ord(code), reason
```
dryajov, line 187: we probably want the string name here, no?

Menduist: Maybe not convinient for unknown codes?

arnetheduck: Above, `RangeError` is a `Defect`  - it cannot be caught reliably - enums must be **manually** checked for range using a case statement

dryajov: Range errors (conceptually) is precisely where making a defect catchable makes sense, in other words is the only excuse to make defect catchable. Now, it should have always been it's own class of errors and not defect, but that's a whole' nother story.

dryajov: Btw I totally agree, defect should be catchable, it just sucks that we can't handle range errors in an uncluttered way.

arnetheduck: >  defect catchable

There's no hard or consistent rule for what should be a defect and what shouldn't, it depends on the context ("in my API, this is a Defect") - conceptually, there's no first principle to lean back on except your mood in the morning when you write the code, so it'll always be prone to bikeshedding - such is the language however, and this along with range checking of integer conversions, null pointer checks and overflows are all things you have to check manually, every time - probably though, it would be trivial to write a macro that verifies the values of an enum and returns an Option or Result - I suspect using such a macro would be even less cluttered than the exception solution.

arnetheduck: @jangko @dryajov @Menduist this btw is something btw that the code needs auditing for across the codebase - I often have to repeat it in reviews which errors are Defects so I suspect there's more of them - `RangeError` for example is an old name, it's called `RangeDefect` in newer version: https://github.com/nim-lang/Nim/blob/v1.4.8/lib/system/exceptions.nim#L156

PR #76, opened 2021-06-25
Fix mutual closing

Menduist:
In the current version, if both parties close the session at the same time, the session get stuck.

### DIFF
--- a/tests/testwebsockets.nim
+++ b/tests/testwebsockets.nim
@@ -456,6 +456,23 @@ suite "Test Closing":
     await session.close()
     check session.readyState == ReadyState.Closed
 
+  test "Mutual closing":
+    proc handle(request: HttpRequest) {.async.} =
+      check request.uri.path == WSPath
+      let server = WSServer.new(protos = ["proto"])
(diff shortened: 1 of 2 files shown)


dryajov:
Ah, nice catch! 

This is how it was originally implemented but it got (wrongly) removed due to failing autobahn tests. We should get rid of `waitForClose` all together as much as possible and only use it where it actually makes sense to simulate some non standard flow, in all other cases we should be closing both ways in our tests.

tests/testwebsockets.nim
```diff
@@ -456,6 +456,23 @@ suite "Test Closing":
     await session.close()
     check session.readyState == ReadyState.Closed
 
+  test "Mutual closing":
+    proc handle(request: HttpRequest) {.async.} =
+      check request.uri.path == WSPath
+      let server = WSServer.new(protos = ["proto"])
```
jangko, line 462: FYI `protos` is optional now, many of old test cases are misleading examples. they actually need no protos  for majority of test cases.

PR #81, opened 2021-07-13
add support for DNS resolution

### LINKED ISSUE #80: client bug: make clear distinction between host address and host name
so far we only test the client using ip address and not using actual host name and then resolve it into address. this become a serious bug when connecting to a real server.

see status-im/nim-json-rpc#109.

example uri: "wss://mainnet.infura.io/ws/v3/infura-id"
host name: "mainnet.infura.io" should be sent to server using "Host" field in HTTP header.

"Host" field cannot use ip address, or get ```Server did not reply with a websocket upgrade: Header code: 403 Header reason: Forbidden Address: 3.85.70.150:443```. It is ok for autobahn because the the autobahn server configuration is using ip address and not using hostname.

host address: need to be resolved using e.g. `getHostByName` / `getAddrInfo`
host port: use default 443 for wss and 80 for ws if the port is not specified.

dryajov:
fix #80 

- reworked API to be more consistent
  - string addresses and Uri types will be now resolved
- made the API more consistent

### DIFF
--- a/examples/autobahn_client.nim
+++ b/examples/autobahn_client.nim
@@ -30,8 +30,7 @@ else:
 
 proc connectServer(path: string, factories: seq[ExtFactory] = @[]): Future[WSSession] {.async.} =
   let ws = await WebSocket.connect(
-    host = "127.0.0.1",
-    port = Port(serverPort),
+    host = "127.0.0.1:$1" % [$serverPort],
     path = path,
     secure=secure,
     flags=clientFlags,
--- a/examples/autobahn_client.nim
+++ b/examples/autobahn_client.nim
@@ -19,12 +19,12 @@ const
 # so we are using different port
 when defined tls:
   const
-    agent      = "websock secure client"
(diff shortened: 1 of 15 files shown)


dryajov:
I'm not sure how best to test this - @jangko @Menduist any ideas?

Menduist:
For reference this is how it's tested on the libp2p side: https://github.com/status-im/nim-libp2p/blob/3fe57f13793adc27aa3e8547825fd8d878d0841a/tests/testnameresolve.nim#L169-L204

For the logic I do simple tests, for the network I mock a dns server we pre-recorded replies.. But this is only possible because I can choose the address of my DNS server.

There is no simple way to resolve with a custom DNS server using `initTAddress`, so the only way I could see in your case is to call another procedure during the tests

dryajov:
Merged - we can address DNS integration testing in a separate PR, right now it seems like this is better than what's already there.

PR #84, opened 2021-07-15
Make HttpServer a case object

dryajov:
Fixes - https://github.com/status-im/nim-libp2p/pull/593

### DIFF
--- a/examples/autobahn_client.nim
+++ b/examples/autobahn_client.nim
@@ -32,8 +32,8 @@ proc connectServer(path: string, factories: seq[ExtFactory] = @[]): Future[WSSes
   let ws = await WebSocket.connect(
     host = "127.0.0.1:$1" % [$serverPort],
     path = path,
-    secure=secure,
-    flags=clientFlags,
+    secure = secure,
+    flags = clientFlags,
     factories = factories
   )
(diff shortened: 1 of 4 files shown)


Menduist (changes requested):
I find this code a bit weird, it's mixing OOP & compile time constructs instead of sticking to just one?

You should either make a single case object and don't extend it, or leave it as before and just replace the
`when server is TlsHttpServer:`
with either a runtime if, or create a `createStream` (or `init` or whatever) method which is extended by the `TlsHttpServer`, no?

EDIT: and in both cases, `proc accept*(server: HttpServer | TlsHttpServer)` shouldn't happen, you should just have `proc accept*(server: HttpServer)`

Menduist (approved):
:+1:

dryajov:
You're right, `HttpServer` shouldn't be extended and `TlsHttpServer` should just be an alias. `accept` is also probably fine taking `HttpServer`. 

The idea behind separating the two has always been more semantic convenience than anything else, it should be clear from calling `TlsHttpServer.create()` that you're creating a Tls server and vice versa.

dryajov:
good catch btw 👍

PR #86, opened 2021-07-27
make distinction between message and data APIs

### LINKED ISSUE #85: Potential bug in UTF-8 validation when the UTF-8 sequence crossing frame boundary
@jlokier pointed out to UTF-8 encoding in websocket spec.

https://datatracker.ietf.org/doc/html/rfc6455#section-5.6

> Text
The "Payload data" is text data encoded as UTF-8.  **Note that a particular text frame might include a partial UTF-8 sequence**; however, the whole message MUST contain valid UTF-8.  Invalid UTF-8 in reassembled messages is handled as described in Section 8.1.

Autobahn have this kind of test cases, but surprisingly, we pass all of them. 
Because we don't handle this case specifically, there is still potential for this bug to manifest itself.

The reason we pass autobahn test is we read using big enough buffer to assemble all arrived UTF-8 frames into complete message and thus satisfy the spec condition. but if we are using smaller buffer, behavior will be different.

dryajov:
fix #85 

- rename `recv` to `recvMsg`
- add `recvTextMsg`/`sendTextMsg` and `recvBinaryMsg`/`sendBinaryMsg`
- move utf8 validation to `recvTextMsg`

### DIFF
--- a/examples/autobahn_client.nim
+++ b/examples/autobahn_client.nim
@@ -80,16 +80,15 @@ proc main() {.async.} =
 
       while ws.readystate != ReadyState.Closed:
         # echo back
-        let data = await ws.recvMsg()
-        let opCode = if ws.binary:
-                       Opcode.Binary
-                     else:
-                       Opcode.Text
+        if ws.binary:
+          let data = await ws.recvBinaryMsg()
+          await ws.sendBinaryMsg(data)
+        else:
+          let data = await ws.recvTextMsg()
+          await ws.sendTextMsg(data)
 
         if ws.readyState == ReadyState.Closed:
           break
-
-        await ws.send(data, opCode)
 
     except WebSocketError as exc:
       error "WebSocket error", exception = exc.msg
--- a/examples/autobahn_client.nim
+++ b/examples/autobahn_client.nim
@@ -43,7 +43,7 @@ proc getCaseCount(): Future[int] {.async.} =
   block:
     try:
       let ws = await connectServer("/getCaseCount")
-      let buff = await ws.recv()
+      let buff = await ws.recvMsg()
       let dataStr = string.fromBytes(buff)
       caseCount = parseInt(dataStr)
       await ws.close()
@@ -60,7 +60,7 @@ proc generateReport() {.async.} =
     trace "request autobahn server to generate report"
     let ws = await connectServer("/updateReports?agent=" & agent)
     while true:
-      let buff = await ws.recv()
+      let buff = await ws.recvMsg()
       if buff.len <= 0:
         break
     await ws.close()
@@ -80,7 +80,7 @@ proc main() {.async.} =
 
       while ws.readystate != ReadyState.Closed:
         # echo back
-        let data = await ws.recv()
+        let data = await ws.recvMsg()
         let opCode = if ws.binary:
                        Opcode.Binary
                      else:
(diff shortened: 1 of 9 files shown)


websock/session.nim
```diff
@@ -433,6 +430,39 @@ proc recv*(
...
 
+proc recvTextMsg*(
+  ws: WSSession,
+  size = WSMaxMessageSize): Future[string] {.async.} =
+  let msg = await ws.recvMsg(size)
+
+  if ws.binary:
+    raise newException(WSInvalidOpcodeError, "Got binary when text expected!")
+
+  if validateUTF8(msg.toOpenArray(0, msg.high)) == false:
```
jangko, line 441: Moving `validateUTF8` to here does not really fix #85, because the `size` parameter can easily trigger the bug if it's value lies in the middle of valid UTF-8 sequence. `validateUTF8` need to be modified into stateful validator. but I guess that can be fixed in next PR.

Menduist:
Maybe a simpler fix would be to buffer the last few bytes if it's not a valid utf8 sequence.
Like, dumb example pseudocode:

```nim
let recvBuffer = recv(requestedSize - textBuffer.len)
msg = textBuffer & recvBuffer

for i in countdown(msg.high, 0):
  if validateUTF8(msg.toOpenArray(0, i)) == true:
    textBuffer = msg[i..^1]
    return msg.toString()
```

Or just remove the `size` parameter from `recvTextMsg`, because as soon as you start talking about size for text, people might expect for the size to be the `strlen`, which in the case of utf8 is != to the byte count.

jangko:
We have alternatives:
1. trim and buffer the last incomplete UTF-8 sequence.
2. read few bytes more to complete the last UTF-8 sequence or error if no more bytes.
3. let user choose 1 or 2

websock/session.nim
dryajov: It doesn't, `size` is the maximum size of the message, if the total amount of bytes read exceeds size it will throw either way. In other words, you can't have partial messages which is what's all of this is about.

dryajov:
> We have alternatives:
> 
> 1. trim and buffer the last incomplete UTF-8 sequence.
> 2. read few bytes more to complete the last UTF-8 sequence or error if no more bytes.
> 3. let user choose 1 or 2

This should not be necessary, the issues as it's described in #85, is due to validation happening on a frame basis, once we move it to a message based level one of three things happen: 
 
1) a message of a valid size (msg.len <= size) with a complete utf8 sequence arrives and we can validate it correctly
2) a message of a valid size (msg.len <= size) with an incorrect utf8 sequence arrives and we fail validation
3) a message of a valid exceeding `size` (msg.len > size) bytes is attempted to be read and we raise an exception `WSMaxMessageSizeError`

It is impossible to read messages that are partial, contain invalid sequences or exceed the max message size, so the issue is mitigated completely. There is no need for stateful readers because the largest valid data unit in websockets is the message.

dryajov:
I think i simple need to rename `size` to `maxMsgSize`, to make this apparent.

examples/autobahn_client.nim
```diff
@@ -80,7 +80,7 @@ proc main() {.async.} =
 
       while ws.readystate != ReadyState.Closed:
         # echo back
-        let data = await ws.recv()
+        let data = await ws.recvMsg()
```
jangko, line 83: I've run the autobahn test locally because the last report was not generated due to some issue in the CI.
the report shows many UTF-8 test is failing. the reason is autobahn test expect us to treat incoming message either as text or binary. but because we are using `recvMsg`, invalid received text message is not verified.

dryajov:
closing in favor of #90

PR #89, opened 2021-07-30
Generalise low-level recv() for accepting partial frames

mjfh:
Intended goal is to make it possible to to verify utf8 text in the WS package. The main showstopper is when frame boundaries are in the middle of utf8 code points. 

There are some more details to consider.
- The internal state machine might allow for frame mode changes beween binary and text (rfc6455 states that all frames have the same type.) This has been solved in a way that text mode is sticky. Switching _binary=>text_ is ok but _text=>binary_ throws an exception (unless confirmed by a descriptor flag.)

Todo:
- require text/binary mode to be all the same in a message once utf8 processing works OK
- add/update unit tests for utf8 scenarios OK

compare against #90, this draft is a study and should be ditched by @dryajov or others after checking usefulness

### DIFF
--- a/websock/session.nim
+++ b/websock/session.nim
@@ -18,6 +18,38 @@ import pkg/chronos/streams/asyncstream
 logScope:
   topics = "websock ws-session"
 
+proc updateReadMode(ws: WSSession): bool =
+  ## This helper function sets text/binary mode for the current frame.
+  ##
+  ## Processing frames might imply a read mode switch. This is typically so
+  ## for the very first frame in binary mode when `ws.binary` has initial
+  ## value.
+  ##
+  ## This function allows to switch from `binary` to `text` mode (aka utf8)
+  ## but not the other way round unless explicitely enabled by setting
+  ## `ws.textSwitchOk` to `true`. Here `text` mode is seen as the more
+  ## restrictive one. Locking `text` mode allows to process the final read
+  ## mode when all of the frame sequence was read after the last frame.
+  ##
+  ## Return value `false` indicates unconfirmed mode switch `text` => `binary`
+
+  if not ws.frame.isNil:
+    # Very first frame, take encoding at face value.
+    if not ws.seen:
+      ws.binary = ws.frame.opcode != Opcode.Text
+      ws.seen = true
+
+    # Accept mode switch from binary => text
+    elif ws.binary and ws.frame.opcode == Opcode.Text:
+      ws.binary = false
+
+    # Beware of a mode switch from text => binary
+    elif not ws.binary and ws.frame.opcode == Opcode.Binary:
+      trace "Read mode changed from text to binary"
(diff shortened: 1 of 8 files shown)


websock/session.nim
```diff
@@ -18,6 +18,38 @@ import pkg/chronos/streams/asyncstream
 logScope:
   topics = "websock ws-session"
 
+proc updateReadMode(ws: WSSession): bool =
```
jangko, line 21: Switching from text to binary or binary to text in a single message are not allowed in the spec. `Opcode.Text` or `Opcode.Binary` should always the be the first frame in a message. In a multi frames message, 2nd frame to final frame should always `Opcode.Cont`.

FYI: the autobahn test result need manual inspection because it will never fail the CI with current setup regardless of failing cases.

mjfh: This was mentioned in the forst item of the introductory comment. The internal state machine _might_ allow for that. So in a defensive manner, an exception is thrown if a state change would not make sense.

mjfh: Re Autobahn: there is no point going there yet until partial frames can be processed reliably (see the new test in testwebsockets.nim.)

mjfh:
> There might be some good cleanup here. But so far, I think this PR misses the point of `recv`, which is a low level primitive that allows reading partial frames (streaming), that should not perform any `utf8` (or otherwise) validation.
> 
> Switching between modes is allowed at message boundaries, but not for frames, so going from text to binary and vice-versa should be fine.

Completely agree, see the coment in the recv(): _utf8 earmarked to go away_ :) Only there is no point in starting something new before the current code is correct. Mainly the _first_ local state variable caused the troubles.

There is only so much i would do at a time.

dryajov:
I took a closer look at this and it makes some unnecessary assumptions about  mode switching between messages, for example there is no need to restrict switching from text to binary or binary to text between messages. 

However, this did identify an issue where mode switching wasn't working properly because the `first` frame wasn't being correctly identified. This was due to a silly attempt to not track an additional `first` flag in the session, this is now rectified in #90. I've moved the test to reproduce this issue to that issue as well and generalized it a bit.

mjfh:
jYou are correct about the mode switching (which was sort of impliciltely supported/handled -- at least unclear to me.) It has been taken care of already as outlined in the general comment. I just had another priority and will com back soon.

dryajov:
I think #90 should address the issues with the general case of UTF-8 handling at message boundaries, there is however an opportunity to add smarter frame streaming for textual data, as pointed out by https://github.com/status-im/nim-websock/issues/85#issuecomment-891027736. We might repurpose this PR for that, WDYT?

mjfh:
Let's talk about this tomorrow. I have a general solution which can be used as a blueprint for more/other ideas ot rejected right away. You can tear it down later. At the moment i would leave it and try how it behaves on the Autobahn.

PR #90, opened 2021-08-03
Perform utf-8 validation at message boundaries

### LINKED ISSUE #85: Potential bug in UTF-8 validation when the UTF-8 sequence crossing frame boundary

dryajov:
Fix #85.

As [section 5.6](https://datatracker.ietf.org/doc/html/rfc6455#section-5.6) of the RFC points, UTF-8 validation has to be performed at message boundaries, frames can carry partial utf-8 sequences:

```
   Text

      The "Payload data" is text data encoded as UTF-8.  Note that a
      particular text frame might include a partial UTF-8 sequence;
      however, the whole message MUST contain valid UTF-8.  Invalid
      UTF-8 in reassembled messages is handled as described in
      Section 8.1.
```

### DIFF
--- a/examples/autobahn_client.nim
+++ b/examples/autobahn_client.nim
@@ -43,7 +43,7 @@ proc getCaseCount(): Future[int] {.async.} =
   block:
     try:
       let ws = await connectServer("/getCaseCount")
-      let buff = await ws.recv()
+      let buff = await ws.recvMsg()
       let dataStr = string.fromBytes(buff)
       caseCount = parseInt(dataStr)
       await ws.close()
@@ -60,7 +60,7 @@ proc generateReport() {.async.} =
     trace "request autobahn server to generate report"
     let ws = await connectServer("/updateReports?agent=" & agent)
     while true:
-      let buff = await ws.recv()
+      let buff = await ws.recvMsg()
       if buff.len <= 0:
         break
     await ws.close()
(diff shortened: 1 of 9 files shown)


examples/autobahn_client.nim
```diff
@@ -80,7 +80,7 @@ proc main() {.async.} =
 
       while ws.readystate != ReadyState.Closed:
         # echo back
-        let data = await ws.recv()
+        let data = await ws.recvMsg()
```
jangko, line 83: This is definitely better than forecast in #86

mjfh (approved):

examples/autobahn_client.nim
mjfh: I will upload the rest of whai i did at #89 which is not unlike the this solution here so @dryajov can use fragments or ditch it all together.

PR #93, opened 2021-09-21
Update asynctest and fix warnings

markspanbroek:
Updates asynctest to version 0.3.0.

This newer version of asynctest has support for unittest2, so the fork in `asyncunit.nim` is no longer necessary and is removed.

GCUnsafe2 warnings are fixed by placing `var` sections under `suite`, so that they are no longer global variables.

Other warnings such as unused imports are also fixed.


arnetheduck:
we collect testing helpers of this kind in https://github.com/status-im/nim-testutils/tree/master/testutils, in general - it would likely be better to reuse that package - in particular, it differentiates the test kind using `asyncTest` instead of `test` which makes it clear that different logic applies

markspanbroek:
Sure, once testutils gains the ability to do async setup and teardown, it should be easy to change the tests to use procSuite, asyncTest, asyncSetup and asyncTeardown from testutils. I personally like the API of asynctest (only change the import), but I can see your point @arnetheduck.

Menduist (approved):
Would be good to unify whatever unit test we use. Since this is a dependency of libp2p and will be installed using `nimble install libp2p`, we should keep the dependency count as low as possible, and when possible, have the same dependencies everywhere (for instance use a single unit test framework for every libp2p dependency & sub dependency)

I'll add that to https://github.com/status-im/nim-libp2p/issues/620, we can merge here in the meantime

Issue #79, opened 2021-07-01
Accept timeout


Menduist:
Yes, by the time we've got to `handleRequest` we've already read all the data (since it's an HTTP request and headers aren't streamed), but we can still have a malicious attacker simply not finishing writing the headers, so I think a timeout when reading headers would be appropriate - this however should go in `nim-ws`.

_Originally posted by @dryajov in https://github.com/status-im/nim-libp2p/pull/593#discussion_r658940877_

arnetheduck:
It's a bit of an open question where timeouts should go, in an `async` world - ie one can put lots of small timeouts everywhere in each step, or a single timeout in the outer, high-level operation which in turn cancels the steps - the latter means that "connect to peer" should take no longer than N seconds, no matter how many steps are involved, and is often appropriate.

arnetheduck:
In particular, imagine setting a 10s timeout on "read bytes from socket" - obviously this doesn't make sense, because a malicious client can trickle bytes one by one effectively disabling the timeout - in such a case, it makes much more sense that "perform handshake" has the timeout.

dryajov:
It makes sense to have a failsafe timeout on `parseRequest`, otherwise an attacker can send a connection but never finish sending the headers.

Doing this at the application level (for example when calling `accept`) is much harder, because you don't really know where in the accept flow it's stuck and it breaks the fundamental assumption with`accept` in that it only returns when it's ready. 

I agree that timeouts are tricky and it's usually better to allow the application to handle that, but in this case it really is very hard to do with the `accept` that this module exposes. 

For example:

```nim
while true:
  let s = await ws.accept().wait(1.seconds)
```

Doesn't make much sense.

dryajov:
The other way around would be to expose `parseRequest`, but it's a bit to extreme IMO.

Menduist:
To add to this, another issue is that on failure, https://github.com/status-im/nim-websock/blob/a697e3585d583ab6b91a159ea7d023461002c927/websock/http/server.nim#L196 can also block the loop for an unknown amount of time (by just using `nc 127.0.0.1 [wss port]`, the request will block the accept loop for the 2 minutes of the header timeout, **and** after that block indefinitively during the closeWait)

rshiv:
should the goal then be to handle the timeout in accept caller , for e.g

https://github.com/status-im/nim-websock/blob/a697e3585d583ab6b91a159ea7d023461002c927/examples/server.nim#L64,

[...]

PR #95, opened 2021-12-10
adds timeout to accept block

rshiv:
This PR addresses issue #79 .

The PR adds timeout to the accept block, which will be used to timeout for async methods, to allow accept block to unblock itself.

The default timeout of 5 seconds is arbitrary, and suggestion for timeout is welcomed.

### DIFF
--- a/websock/http/server.nim
+++ b/websock/http/server.nim
@@ -162,7 +162,7 @@ proc handleTlsConnCb(
   finally:
     await stream.closeWait()
 
-proc accept*(server: HttpServer): Future[HttpRequest]
+proc accept*(server: HttpServer, timeout: Duration = 5.seconds): Future[HttpRequest]
   {.async, raises: [Defect, HttpError].} =
 
   if not isNil(server.handler):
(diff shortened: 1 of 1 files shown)


websock/http/server.nim
```diff
@@ -162,7 +162,7 @@ proc handleTlsConnCb(
   finally:
     await stream.closeWait()
 
-proc accept*(server: HttpServer): Future[HttpRequest]
+proc accept*(server: HttpServer, timeout: Duration = 5.seconds): Future[HttpRequest]
```
dryajov, line 165: Accept shouldn't take the timeout, it's better to allow configuring it during object creation and we should call it something like `HandshakeTimeout`.

rshiv: resolved in a6ba0b9256ef47ba4423441949c01b2be270ea32

websock/http/server.nim
```diff
@@ -191,22 +192,36 @@ proc accept*(server: HttpServer): Future[HttpRequest]
 
   trace "Got new request", isTls = server.secure
   try:
-    return await server.parseRequest(stream)
+    let parsedRequestfut = server.parseRequest(stream)
+    let res = await withTimeout(parsedRequestfut, server.handshakeTimeout)
+    if not res:
+      # Timeout
+      raise newException(HttpError, "Didn't Parse request in time!")
```
dryajov, line 199: I'd rather we don't raise multiple times in this way. Can we restructure this so that we only raise once with the appropriate error?

Menduist:
Closed in favor of #102

PR #97, opened 2021-12-16
Remove version pin on chronicles

markspanbroek:
Caused dependency conflicts when combining websock with other
packages that depend on chronicles, now that chronicles has
a new version (0.10.2).


Menduist:
Could we have the >= at least? It was required for isLogFormatUsed iirc

Menduist (approved):

PR #100, opened 2022-01-25
avoid spin/live-lock on 0 bytes read

### LINKED ISSUE #94: Nimbus livelock in recvMsg
https://gist.github.com/stefantalpalaru/311ea3e48f104250c6c3dbc4ee0b3718

dryajov:
fix #94


Menduist:
That should fix it, but should we just ignore it like here, or raise something?

Raising something would close the connection, which might be the right thing to do

dryajov:
> That should fix it, but should we just ignore it like here, or raise something?
> 
> Raising something would close the connection, which might be the right thing to do

This fix cover the case of empty payload frames, for other types of errors `recv` already raises an exception which we would propagate accordingly.

Menduist:
If there was a legit empty frame, it would read it, return, and read (giving back control) a new one, until there is a non-empty one, no?

So it should work as usual, not live-lock? Or I missed something?

My understanding is the live-lock happens when the connection is badly closed or other weird stuff

Menduist:
I mean, by returning it, you move the issue to the caller, which will livelock if he does something like:
```nim
while something:
  process(await conn.recvMsg())
```

dryajov:
> My understanding is the live-lock happens when the connection is badly closed or other weird stuff

You can send empty payloads without signaling a finished frame, which is totally valid btw.

Menduist:
> You can send empty payloads without signaling a finished frame, which is totally valid btw.

Yes, but that wouldn't cause a livelock, would it?

dryajov:
most likely addressed by #109

PR #102, opened 2022-02-28
Accept timeout

### LINKED ISSUE #79: Accept timeout

Menduist:
Closes #79 

Ended up doing something very similar to #95, because each step of the parseRequest (including it's error handling) can go stale

### DIFF
--- a/websock/http/server.nim
+++ b/websock/http/server.nim
@@ -200,7 +200,7 @@ proc accept*(server: HttpServer): Future[HttpRequest]
     raise newException(HttpError, "Timed out parsing request")
   except CatchableError as exc:
     # Can't hold up the accept loop
-    asyncSpawn stream.closeWait()
+    stream.close()
     raise exc
 
 
--- a/websock/http/server.nim
+++ b/websock/http/server.nim
@@ -192,8 +192,11 @@ proc accept*(server: HttpServer): Future[HttpRequest]
 
   trace "Got new request", isTls = server.secure
   try:
-    let parseFut = server.parseRequest(stream)
-    if await withTimeout(parseFut, server.handshakeTimeout + 2.seconds):
+    let
+      parseFut = server.parseRequest(stream)
+      # Failsafe at timeout * 1.05
+      timeout = server.handshakeTimeout + server.handshakeTimeout div 20
+    if await withTimeout(parseFut, timeout):
       return parseFut.read()
     raise newException(HttpError, "Timed out parsing request")
   except CatchableError as exc:
--- a/websock/http/server.nim
+++ b/websock/http/server.nim
@@ -29,6 +29,7 @@ type
 
   HttpServer* = ref object of StreamServer
     handler*: HttpAsyncCallback
+    handshakeTimeout*: Duration
     case secure*: bool:
     of true:
       tlsFlags*: set[TLSFlags]
@@ -72,7 +73,7 @@ proc parseRequest(
   try:
     let hlenfut = stream.reader.readUntil(
       addr buffer[0], MaxHttpHeadersSize, sep = HeaderSep)
-    let ores = await withTimeout(hlenfut, HttpHeadersTimeout)
+    let ores = await withTimeout(hlenfut, server.handshakeTimeout)
     if not ores:
       # Timeout
       trace "Timeout expired while receiving headers", address = $remoteAddr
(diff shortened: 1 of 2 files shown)


websock/http/server.nim
```diff
@@ -191,22 +192,30 @@ proc accept*(server: HttpServer): Future[HttpRequest]
 
   trace "Got new request", isTls = server.secure
   try:
-    return await server.parseRequest(stream)
+    let
+      parseFut = server.parseRequest(stream)
+      # Failsafe at timeout * 1.05
+      timeout = server.handshakeTimeout + server.handshakeTimeout div 20
```
dryajov, line 198: I'm not sure this is the right approach, why would we have two timeouts one for the initial request in `parseRequest` another here, does it make sense to wait twice, whats the reasoning? I would really just stick with one timeout on `parseRequest().wait()` (or similar), wether it took too long to send the headers or during validation is irrelevant, IMO.

dryajov: I see, ok that makes sense then. In that case I would separate the two timeouts so that we can tweak them separately. For (contrived) example, we might know that a particular implementation is slow sending headers and adjust for that. We can either have them as two separate options - `headersTimeout` and `handshakeTimeout`, or amortize the `headersTimeout` into the `handshakeTimeout`, i.e. `handshakeTimeout + handshakeTimeout`.

The reason is because otherwise you're making arbitrary choices - why `handshakeTimeout + handshakeTimeout / 20` and not `handshakeTimeout + handshakeTimeout / 50`, this makes it difficult for the users/consumers to reason about it. Giving the user the ability to tweak this options makes the usage of the library more predictable and removes the need to for us to make this decisions for the user.

websock/http/server.nim
```diff
@@ -207,18 +206,34 @@ proc accept*(server: HttpServer): Future[HttpRequest]
...
   ## Make a new HTTP Server
   ##
 
-  var server = HttpServer(handler: handler, handshakeTimeout: handshakeTimeout)
+  var server = HttpServer(
+    handler: handler,
+    headersTimeout: headersTimeout,
+    handshakeTimeout:
+      if handshakeTimeout == 0.seconds:
+        headersTimeout + headersTimeout div 20
```
dryajov, line 224: I still don't get why we need to divide by 20.

Menduist: It's just a sane default, handshakeTimeout = headersTimeout * 1.05
The div 20 is just to stay in integer world, which is generally better

Having headersTimeout * 2 is too big imo, that basically means we wait for headersTimeout to read and headers, and then leave headersTimeout again to send the error. Too long, imo

websock/http/server.nim
```diff
@@ -191,22 +192,30 @@ proc accept*(server: HttpServer): Future[HttpRequest]
...
+      parseFut = server.parseRequest(stream)
+      # Failsafe at timeout * 1.05
+      timeout = server.handshakeTimeout + server.handshakeTimeout div 20
+    if await withTimeout(parseFut, timeout):
+      return parseFut.read()
+    raise newException(HttpError, "Timed out parsing request")
   except CatchableError as exc:
-    await stream.closeWait()
+    # Can't hold up the accept loop
+    asyncSpawn stream.closeWait()
```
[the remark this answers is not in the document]
dryajov: I think it's better, just for consistency's sake, plus `asyncCheck` is something to avoid, unless absolutely necessary.

websock/http/server.nim
dryajov: OK, I'm not opposed to the argument, I guess I'm missing where does the `1.05` comes from - are you talking about operator precedence `headersTimeout + (headersTimeout div 20)`? I guess I'm missing something here.

websock/http/server.nim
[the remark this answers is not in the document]
Menduist: Done

websock/http/server.nim
Menduist:
```suggestion
        # default to headersTimeout * 1.05
        headersTimeout + (headersTimeout div 20)
```

dryajov (approved):

PR #103, opened 2022-03-03
add http header plugin

### LINKED ISSUE #101: jwt authentication for rpc interface
The upcoming `engine` JSON-RPC interface, exposed by eth-EL and consumed by eth-CL, needs to be authenticated.

Read the rest of this information in:
- [Eth-Engine API authentication spec](https://github.com/ethereum/execution-apis/blob/main/src/engine/authentication.md)
- https://jwt.io/
- [RFC 7519](https://datatracker.ietf.org/doc/html/rfc7519)

The authentication is done when client perform HTTP upgrade request, not during websocket handshake.
Maybe we can implement some sort of HTTP authentication plugin(not to be confused with websocket plugin).

we can implement this as HTTP header validation plugin, thus we don't need to deal with the jwt itself but only care about the plugin result.

jangko:
fix #101

### DIFF
--- a/tests/testhooks.nim
+++ b/tests/testhooks.nim
@@ -1,0 +1,201 @@
+## nim-websock
+## Copyright (c) 2021 Status Research & Development GmbH
+## Licensed under either of
+##  * Apache License, version 2.0, ([LICENSE-APACHE](LICENSE-APACHE))
+##  * MIT license ([LICENSE-MIT](LICENSE-MIT))
+## at your option.
+## This file may not be copied, modified, or distributed except according to
+## those terms.
+
+import std/strutils
+import pkg/[
+  httputils,
+  chronos,
+  chronicles,
+  stew/byteutils,
+  asynctest/unittest2]
+
+import ../websock/websock
+
+import ./helpers
+
+let address = initTAddress("127.0.0.1:8888")
+
+type
+  TokenHook* = ref object of Hook
+    status: int
+    token: string
+    request: HttpRequest
+
+proc clientAppendGoodToken(ctx: Hook, headers: var HttpTable):
+     Result[void, string] {.gcsafe, raises: [Defect].} =
+  headers.add("auth-token", "good-token")
+  return ok()
+
+proc clientAppendBadToken(ctx: Hook, headers: var HttpTable):
+     Result[void, string] {.gcsafe, raises: [Defect].} =
+  headers.add("auth-token", "bad-token")
+  return ok()
+
+proc clientVerify(ctx: Hook, headers: HttpTable):
+     Future[Result[void, string]] {.async, gcsafe, raises: [Defect].} =
+  var p = TokenHook(ctx)
+  p.token = headers.getString("auth-status")
+  return ok()
+
+proc serverVerify(ctx: Hook, headers: HttpTable):
+     Future[Result[void, string]] {.async, gcsafe, raises: [Defect].} =
+  var p = TokenHook(ctx)
+  if headers.getString("auth-token") == "good-token":
+    p.status = 101
+  return ok()
+
+proc serverAppend(ctx: Hook, headers: var HttpTable):
+     Result[void, string] {.gcsafe, raises: [Defect].} =
+  var p = TokenHook(ctx)
+  if p.status == 101:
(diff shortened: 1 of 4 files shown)


dryajov:
Looks reasonable, I'll take a closer look ASAP.

websock/types.nim
```diff
@@ -112,6 +115,26 @@ type
...
+  # server exec order:
+  #   1. verify request header
+  #   2. append to response header
+  # ------------------------------
+  # Handshake exec order:
+  # 1. client append to request header
+  # 2. server verify request header
+  # 3. server reply with response header
+  # 4. client verify response header from server
+  HdrPlugin* = ref object of RootObj
```
dryajov, line 130: I'd say that `plugin` is a bit misleading as it conflicts with native websocket plugins nomenclature - how about `hook`?

jangko: no problem. `hook` is more ergonomic and easier to type.

But I have other problem during the second step:
```text
1. client append to request header
2. server verify request header                    # <---- this second step
3. server reply with response header
4. client verify response header from server
```

There are cases where authentication not replied with `Http101 + response header` but `Http401` or other Http Error code. Still thinking the best approach to this situation.

dryajov: > There are cases where authentication not replied with Http101 + response header but Http401 or other Http Error code. Still thinking the best approach to this situation.

What specifically is making handing this case difficult? Is it propagating the error to the caller or something else I'm missing?

tests/testhooks.nim
```diff
@@ -77,12 +77,26 @@ proc serverHook(): Hook =
...
   )
 
+proc serverVerifyWithCode(ctx: Hook, headers: HttpTable):
+     Result[HttpCode, string] {.gcsafe, raises: [Defect].} =
+  var p = TokenHook(ctx)
+  if headers.getString("auth-token") == "good-token":
+    p.status = 101
+    ok(Http101)
+  else:
+    ok(Http401) # Unauthorized
```
jangko, line 87: To illustrate the case, I've made an example returning HttpCode from hook `verify` result. but It feels weird.
```nim
  Hook* = ref object of RootObj
    append*: proc(ctx: Hook,
                  headers: var HttpTable): Result[void, string]
                  {.gcsafe, raises: [Defect].}
    verify*: proc(ctx: Hook,
                  headers: HttpTable): Result[void, string]
                  headers: HttpTable): Result[HttpCode, string]     # weird
                  {.gcsafe, raises: [Defect].}
```

Second approach also not very intuitive:
```Nim
Hook* = ref object of RootObj
    append*: proc(ctx: Hook,
                  headers: var HttpTable): Result[void, string]
                  {.gcsafe, raises: [Defect].}
    verify*: proc(ctx: Hook,
                  headers: HttpTable): Result[void, string]
                  headers: HttpTable): Result[void, (HttpCode, string)]  # also not very good
                  {.gcsafe, raises: [Defect].}
```

Third approach is less descriptive:
```Nim
Hook* = ref object of RootObj
    append*: proc(ctx: Hook,
                  headers: var HttpTable): Result[void, string]
                  {.gcsafe, raises: [Defect].}
    verify*: proc(ctx: Hook,
                  headers: HttpTable): Result[void, string]
                  headers: HttpTable): Result[void, HttpCode]  # limiting the error message to HttpCode
                  {.gcsafe, raises: [Defect].}
```

dryajov: I think that there is no need to communicate anything back to the caller of the hook because we aren't doing anything with it at all... 

The hooks are application specific and executed within the application context they where declared in. Anything application specific related to the failed verification should happen in the hook itself. The only reasonable return value from the hook is a straight forward `bool` that indicates whether connection establishing should proceed or halt.

dryajov: For logging purposes, we might return a `string`, just as in one of your examples.

jangko: okay. I think I get the idea. Will clean it up and add some docs.

dryajov: I think this `Result[void, string]` would be a reasonable return value, given that we want a string for logging purposes.

dryajov (approved):
LGTM

jangko:
rebase and merged

PR #109, opened 2022-05-17
Fix some issues with recv hanging

### LINKED ISSUE #94: Nimbus livelock in recvMsg

Menduist:
Hopefully, closes #94

### DIFF
--- a/websock/frame.nim
+++ b/websock/frame.nim
@@ -205,6 +205,9 @@ proc decode*(
 
   frame.length = finalLen
 
+  if frame.length > WSMaxMessageSize:
+    raise newException(WSPayloadLengthError, "Frame too big: " & $frame.length)
+
   # Do we need to apply mask?
   frame.mask = (b1 and 0x80) == 0x80
   if masked == frame.mask:
--- a/websock/session.nim
+++ b/websock/session.nim
@@ -312,6 +312,10 @@ proc recv*(
   ## Use this to stream data from frames
   ##
 
+  assert ws.reading == false, "Only one concurrent read allowed"
+  ws.reading = true
+  defer: ws.reading = false
+
   var consumed = 0
   var pbuffer = cast[ptr UncheckedArray[byte]](data)
   try:
--- a/websock/session.nim
+++ b/websock/session.nim
@@ -312,7 +312,7 @@ proc recv*(
   ## Use this to stream data from frames
   ##
 
-  assert ws.reading == false, "Only one concurrent read allowed"
+  doAssert ws.reading == false, "Only one concurrent read allowed"
   ws.reading = true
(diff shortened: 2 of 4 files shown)


websock/session.nim
```diff
@@ -347,8 +352,8 @@ proc recv*(
         trace "Reading bytes from frame stream", len
         let read = await ws.frame.read(ws.stream.reader, addr pbuffer[consumed], len)
         if read <= 0:
-          trace "Didn't read any bytes, breaking"
-          break
+          trace "Didn't read any bytes, stopping"
+          raise newException(WSClosedError, "WebSocket is closed!")
```
Menduist, line 356: read = 0 means EOF

websock/frame.nim
```diff
@@ -205,6 +205,9 @@ proc decode*(
 
   frame.length = finalLen
 
+  if frame.length > WSMaxMessageSize:
```
Menduist, line 208: A too big frame.length (of instance 13402712495349628929) causes issues in recv, capping it here

dryajov:
Also, are you fuzing it to get this failures or doing something else?

websock/session.nim
arnetheduck: the point with returning 0 on EOF is that exceptions are generally not suitable for entirely normal behavior - ie this way, a 100% correct application that closes the stream regularly cannot work without at least one exception being fired - I'd generally expect this of any `recv`-style function (recv itself in bsd sockets works this way - it returns 0 on EOF and -1 on errors)

arnetheduck: (to be clear, this `recv` function should also return 0 on a regular stream closure)

Menduist: Like often, the line between "regular behavior" and "exceptional behavior" gets fuzzy
Here for instance, this is not a regular behavior, because the client is closing the connection "unsafely", in the middle of the frame

More importantly, websocket allows empty frame, which would "legally" return 0 from `recvMsg`
So unfortunately, from this procedure, the only way I currently have to tell `recvMsg` that the stream is closed is through an exception

arnetheduck: > Here for instance, this is not a regular behavior, because the client is closing the connection "unsafely", in the middle of the frame

ah, fair point then - in general cases, it's not much of a "fuzzy" border for `recv` - it's more of a convention that's quite common (which `readOnce` generally continues, as a tradition), that closing is not considered an error at all (regardless of error reporting method) if it's done through a "regular" `close`

Menduist:
Yes fuzzing only, I'll share my setup & tests soon

Menduist:
Autobahn passed, merging

PR #125, opened 2022-09-01
allow passing extra client headers

etan-status:
`nim-websock` currently lacks a way to pass additional client headers
such as `Authorization` and `Origin` which are needed for `nimbus-eth2`.
Adding an extra argument for those opens up swapping from `news`.


etan-status:
`macos-amd64 (Nim devel)` unrelated fail due to `could not load: libcrypto.1.1.dylib`.
Nim devel on macOS temporarily needs `brew install openssl@1.1`, probably fine to ignore for this PR.

etan-status:
Ah, just realized, the same functionality could be reached by using the `hook` system.

PR #126, opened 2022-09-02
properly support concurrent message sending

etan-status:
`nim-websock` suffered from a number of issues that are being addressed:

1. Long messages > `frameSize` (default 1 MB) were split into fragments
   of `frameSize` each. However, when a concurrent message is sent, it
   may be interleaved among the fragments of an already-sending message.
   This is only allowed for control packets without a mux extension.

2. When the WebSocket session is closed, a msg may have been partially
   received. This partial frame was reported as a full message, without
   indication that the receiving was canceled. This behaviour is fixed
   by raising a `WSClosedError` instead of reporting the partial msg.

3. When an individual `send` operation was canceled, it would actually
   stop sending the remainder of a potentially partially sent messages.
   This would corrupt the stream for concurrent and followup operations.
   Cancellation is now inhibited for the message currently sending.
   It is still possible to cancel messages that are not yet scheduled.

4. Messages could get reordered when using asynchronous encoders. This
   is addressed by delaying followup messages until the current message
   is fully encoded and transmitted (except for control packets).

### DIFF
--- a/websock/session.nim
+++ b/websock/session.nim
@@ -96,7 +96,7 @@ proc nonCancellableSend(
   ws: WSSession,
   data: seq[byte] = @[],
   opcode: Opcode): Future[void]
-  {.async, raises: [Defect, WSClosedError].} =
+  {.async.} =
   ## Send a frame
   ##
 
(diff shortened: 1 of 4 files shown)


websock/session.nim
```diff
@@ -116,13 +119,86 @@ proc send*(
...
+    if fut.finished:
+      return
+    if sendFut.failed:
+      fut.fail(sendFut.error)
+    elif sendFut.cancelled:
+      fut.cancel()
+    else:
+      fut.complete()
+
+  sendFut.addCallback(handleSent)
```
Menduist, line 150: This style of async (callback) is generally discouraged, especially in this case where you can just
```nim
await sendFut # maybe wrap in try

# move handleSent content here
```

websock/session.nim
```diff
@@ -116,13 +119,86 @@ proc send*(
...
+
+  if opcode.isControl:
+    # Control frames (see Section 5.5) MAY be injected in the middle of
+    # a fragmented message.  Control frames themselves MUST NOT be
+    # fragmented.
+    # See RFC 6455 Section 5.4 Fragmentation
+    discard ws.doSend(data, opcode, fut)
+    return fut
+
+  if ws.sendFut != nil:
```
Menduist, line 190: Might be easier to do `if ws.sendFut != nil or ws.sendFut.finished` (not need to set it to nil afterwards)

etan-status: Setting it to `nil` releases associated memory though.

etan-status: And when `ws.sendFut.finished`, it may still be necessary to go through `addLast` to ensure that potential continuations in the queue are also scheduled.

websock/session.nim
etan-status: The goal here is to unlink `fut` from `sendfut`, so that `sendFut` does not get cancelled when caller cancels `fut`.

With `await` syntax it would not really get clearer, as the result still needs to be propagated from `sendFut` to `fut`, in any case. There's also the early exit `if fut.cancelled`, which is tricky to implement from `async` function (which this is not), as `newCancelledError` is not exported.

Menduist: Indeed currently it's a bit of a pain to do this, though it will improve with a `awaitnc` or similar
Anyway, here is a branch with saner async usage, imo: https://github.com/status-im/nim-websock/commit/a4263867cc42903d64eaa7ccbba343cac22af38c

websock/session.nim
Menduist: > Setting it to nil releases associated memory though.

When finished, a Future will already release the closure environment on it's own: https://github.com/status-im/nim-chronos/blob/3e4b47e60a6033ec6732479b308982648ebe6074/chronos/asyncfutures2.nim#L373

Leaving only a few bytes behind

websock/session.nim
etan-status: OK, I'm not sure how the performance characteristics change with `try` / `except` here (seems reasonable as a 1 per message overhead), and it's a bit less defensive in a way that it translates `CancelledError` into `fail()` instead of `cancel()` and does not deal with client messing with returned fut in other ways than running `cancel()` on it, but generally it still looks correct to me when used correctly. Will give it a shot as part of https://github.com/status-im/nimbus-eth2/pull/4061

websock/session.nim
etan-status: Ah, that's good to know. Thanks for the pointer!

etan-status:
Also related to `nimbus-eth2` usage: https://github.com/status-im/nim-json-rpc/pull/146

websock/session.nim
Menduist: > it's a bit less defensive in a way that it translates CancelledError into fail() instead of cancel()

Unless I'm missing something, these can't be cancelled anyway, so CancelledError shouldn't happen?

> does not deal with client messing with returned fut in other ways than running `cancel()` on it

imo the only two possible operations are `await` it or `cancel` it, is there another operation you would expect?

etan-status: When normally used by a caller, that's correct. But it's a library, so there's nothing preventing API misuse. There's also this little extra mental step when reviewing to see "oh _that's_ why it's fine to not propagate cancellation". It's fine by me either way, under normal operation it should work equally well.

Menduist: Right, only `send` is exported, so it shouldn't be possible to misuse as a lib, but ofc we need to be careful to keep it sensible here. Hopefully we'll have `awaitnc` sooner than later :)

etan-status: The `retFut` is also exported, client could `fail(customError)` on it, then we crash because we try to complete it once more. And the `sendFut` is exported via types module. But agree, it's alright for now.

Menduist: Good points!

websock/session.nim
```diff
@@ -116,13 +119,61 @@ proc send*(
...
+      retFut.fail(exc)
+
+  asyncSpawn handleSend()
+  retFut
+
+proc sendLoop(ws: WSSession) {.gcsafe, async.} =
+  while ws.sendQueue.len > 0:
+    let task = ws.sendQueue.popFirst()
+    if task.fut.cancelled:
+      continue
```
etan-status, line 150: only checking for `cancelled` to behave similarly in case that client abuses API and completes/fails in advance, as if the queue was not used

etan-status:
I did prolonged manual testing on `nimbus-eth2` as part of https://github.com/status-im/nimbus-eth2/pull/4061 and `nim-websock` was very reliable.

Menduist (approved):

PR #136, opened 2023-01-25
Remove recvMsg pre-allocation

Menduist:
Alternative to #135


websock/session.nim
```diff
@@ -349,7 +349,7 @@ proc ping*(
 
 proc recv*(
   ws: WSSession,
-  data: pointer | ref seq[byte],
+  data: pointer | ptr byte | ref seq[byte], # nim bug: pointer doesn't match ptr byte?
```
Menduist, line 352: https://github.com/nim-lang/Nim/issues/21298

PR #139, opened 2023-05-10
Free send queue on session closing

Menduist:


Menduist:
This isn't actually required, nevermind

PR #148, opened 2023-06-19
replace `utils` module with direct `generate` calls

etan-status:
The `utils` module uses a convoluted mechanism for filling byte arrays with random data. The `generate` function can be used directly, making the `utils` module obsolete.


dryajov (approved):
LGTM

PR #165, opened 2024-10-31
chore: use supported selector operators

emizzle:
Use supported nimble <0.16.2 selector operators when specifying dependency versions. Nimble ≥0.16.1 requires nim 2.0, so this is (a more explicit) workaround for projects/libs still using 1.6.


arnetheduck:
alright, let's roll with this for now to unblock, though the long-term plan remains that users have a newer nimble and update it independently of their nim version

PR #173, opened 2025-09-15
Fix query truncation in WebSocket.connect

darkestpigeon:
When `WebSocket.connect` is called with an `Uri` object, the query part in completely ignored. Fixing this.

### DIFF
--- a/websock/websock.nim
+++ b/websock/websock.nim
@@ -229,9 +229,15 @@ proc connect*(
   if uri.port.len <= 0:
     uri.port = if secure: "443" else: "80"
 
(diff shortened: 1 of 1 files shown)


websock/websock.nim
```diff
@@ -229,9 +229,15 @@ proc connect*(
   if uri.port.len <= 0:
     uri.port = if secure: "443" else: "80"
 
+  let pathQuery =
```
arnetheduck, line 232: would `$uri` work here? It should reconstruct the whole uri (including logins etc)

PR #181, opened 2025-12-05
feat(wstransport): add support for concurrent accept

fcecin:
**Overview**

This PR helps address https://github.com/logos-messaging/nwaku/issues/3634 by adding public APIs `acceptStream` and `processHttpRequest` to `HttpServer`. 

Also added a "Test accept" test suite which tests both `accept` and `acceptStream` with `processHttpRequest`.

**Rationale**

Client code that doesn't tolerate the head-of-line blocking of the existing `accept` API can instead (i) call `acceptStream` in a loop, in a dedicated task to create remote-client sockets from the server socket, and (ii) spawn new tasks to call `processHttpRequest` and wait for and process HTTP headers arriving from the client socket (which can take a long time and shouldn't block new remote-client sockets).

Alternatives (tried before):

* [Push everything into nim-websock](https://github.com/status-im/nim-websock/pull/179): The most transparent and localized solution to the bug, but it is arguably the ugliest solution; out of nim-websock's scope; not ideal.
* [Spawn async tasks on the client to call the existing `accept`](https://github.com/vacp2p/nim-libp2p/pull/1919): Even with a [fix to Chronos' inability to handle StreamServer.accept multitasking](https://github.com/status-im/nim-websock/pull/180) we have observed [very weird behavior](https://github.com/vacp2p/nim-libp2p/pull/1919#discussion_r2572290148); in any case, this kind of solution [fights Chronos](https://github.com/status-im/nim-websock/pull/180#issuecomment-3602313319) and is not ideal.
* Spawn an HTTP header parsing task inside `accept`: even if it's still nominally just an `accept` API, client code still has to asyncSpawn, in addition to the asyncSpawn inside `accept`. `HttpRequest` object becomes "hollow" and needs to be populated later; clients need to know that a future inside `HttpRequest` must be awaited before accessing its data fields.
* Callback/handlers: will be deprecated from nim-websock; doesn't solve the problem of the client still having to duplicate HTTP header handling code.

### DIFF
--- a/tests/testwebsockets.nim
+++ b/tests/testwebsockets.nim
@@ -1193,6 +1193,16 @@ suite "Partial frames":
     await lowLevelRecv(7, 10, 5)
 
 suite "Test accept":
+  setup:
+    var
+      server = createServerNoStart(address)
+      wsServer = WSServer.new()
+
+  teardown:
+    if server != nil:
+      server.stop()
+      waitFor server.closeWait()
+
   proc runEchoClient(address: TransportAddress, id: string): Future[void] {.async.} =
     let client = await connectClient(address).wait(2.seconds)
     try:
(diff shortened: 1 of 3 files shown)


tests/testwebsockets.nim
```diff
@@ -1191,3 +1191,117 @@ suite "Partial frames":
...
+    await allFutures(clientFuts).wait(5.seconds)
+
+  template withTestServer(body: untyped) =
+    let server {.inject.} = createServerNoStart(address)
+    let wsServer {.inject.} = WSServer.new()
+    try:
+      body
+    finally:
+      server.stop()
+      await server.closeWait()
```
Ivansete-status: Maybe we could use `setup` and `teardown` instead. Similarly on how we use it in `logos-messaging-nim`.

tests/testwebsockets.nim
```diff
@@ -1191,3 +1191,117 @@ suite "Partial frames":
 
   asyncTest "receiver frameSize greater than sender":
     await lowLevelRecv(7, 10, 5)
+
+suite "Test accept":
+  proc runEchoClient(address: TransportAddress, id: string): Future[void] {.async.} =
+    let client = await connectClient(address).wait(2.seconds)
+    try:
+      let payload = "client-" & id
+      await client.send(payload)
```
Ivansete-status: It is interesting to see how this proceeds further. Maybe it waits until the payload is put into a kernel's buffer?

websock/http/server.nim
```diff
@@ -145,6 +150,13 @@ proc accept*(server: HttpServer): Future[HttpRequest] {.async.} =
       raise (ref HttpError)(msg: error)
 
   trace "Got new request", isTls = server.secure
+  return stream
+
+# TODO async raises not implemented for accept because it breaks libp2p (1.13.0
+#      at the time of writing)
```
Ivansete-status: What does this comment mean :) ?

websock/http/server.nim
```diff
@@ -129,9 +136,7 @@ proc handleConnCb(
   finally:
     await stream.closeWait()
 
-# TODO async raises not implemented for accept because it breaks libp2p (1.13.0
-#      at the time of writing)
-proc accept*(server: HttpServer): Future[HttpRequest] {.async.} =
+proc acceptStream*(server: HttpServer): Future[AsyncStream] {.async.} =
```
Ivansete-status: May add a tinny comment under the signature?

websock/http/server.nim
```diff
@@ -86,6 +86,13 @@ proc readHttpRequest(
     headers: request.toHttpTable(), stream: stream, uri: request.uri().parseUri()
   )
 
+proc processHttpRequest*(
```
Ivansete-status: Kindly add a tiny comment, below signature, elaborating

Ivansete-status (approved):
LGTM! Thanks so much for it indeed! 🙌 
I'm just adding some nitpick comments that I hope you find useful.

fcecin:
Superseded by #193

PR #185, opened 2026-03-31
version 0.2.2

tersec:


Ivansete-status (approved):
LGTM! Thanks for it! 🙌

arnetheduck:
what about https://github.com/status-im/nim-websock/pull/184 ?

arnetheduck:
ie this was probably a mistake semver-wise - https://github.com/status-im/nim-websock/commit/5e9bcac3257afc53e91bf4d79e9804eafe0728d0 is a major rewrite

tersec:
> ie this was probably a mistake semver-wise - [5e9bcac](https://github.com/status-im/nim-websock/commit/5e9bcac3257afc53e91bf4d79e9804eafe0728d0) is a major rewrite

https://semver.org/
> Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.

There is no semver here.

arnetheduck:
> There is no semver here.

read on:

```
How should I deal with revisions in the 0.y.z initial development phase?

The simplest thing to do is start your initial development release at 0.1.0 and then increment the minor version for each subsequent release.
```

"release" here is usually read as "sufficiently large changes to motivate a release" in my experience (while patch releases are bugfixes as usual), ie as a courtesy to consumers more than anything else - the spirit of semver is avoid downstream breakage, and this is one useful trick/soft-rule.

arnetheduck:
This release also breaks downstream packages because it drops 1.6 support without updating the nimble `requires` - ie the release is not being tested with 1.6 but still claims support

arnetheduck:
https://github.com/status-im/nim-json-rpc/pull/266#discussion_r3063912870 - cc @narimiran this is the root cause of the issue, ie borked and misnumbered websock release.

https://github.com/status-im/nim-json-rpc/blob/7434eaac677939984639b85dbe17afe557d6c45e/json_rpc.nimble#L27 is illustrative of how this misnumbered release causes unnecessary breakage - tagged 0.3, it would not have been selected for json_rpc

arnetheduck:
So to fix this mess, the easiest thing is actually to retract the release (which is done by simply removing the tag).

Then we can update the nimble file and publish this release as 0.3.0.

tersec:
Tag removed.

PR #190, opened 2026-04-10
remove Nim 1.6 support

tersec:
https://github.com/status-im/nim-json-rpc/pull/266#discussion_r3063912870
> this is a bug that should be fixed in websock 0.2.2 - it's incorrectly claiming 1.6 support which makes nimble choose the wrong version here (it should have gracefully degraded to 0.2.1 for example) - changing the limit here does not actually fix the underlying issue.

### DIFF
--- a/websock.nimble
+++ b/websock.nimble
@@ -14,7 +14,7 @@ description = "WS protocol implementation"
 license     = "MIT"
 skipDirs    = @["examples", "tests"]
 
-requires "nim >= 1.6.0"
+requires "nim >= 2.0.16"
 requires "chronos >= 4.2.0 & < 4.4.0"
 requires "httputils >= 0.2.0"
 requires "chronicles >= 0.10.2"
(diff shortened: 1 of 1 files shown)


websock.nimble
```diff
@@ -25,26 +25,23 @@ requires "results"
 requires "zlib"
 
 task test, "run tests":
-  let
-    envNimflags = getEnv("NIMFLAGS")
-    nimFlags = envNimFlags &
-      " --verbosity:0 --styleCheck:usages --styleCheck:error" &
-      " --mm:refc"
```
Ivansete-status, line 32: I think is interesting to keep `mm` explicit

websock.nimble
```diff
@@ -14,7 +14,7 @@ description = "WS protocol implementation"
 license     = "MIT"
 skipDirs    = @["examples", "tests"]
 
-requires "nim >= 1.6.0"
+requires "nim >= 2.0.16"
```
Ivansete-status, line 17: fyi we use 2.2.4 in `logos-delivery`

Ivansete-status (approved):

websock.nimble
tersec: Yeah, that was an incomplete change, see https://github.com/status-im/nim-websock/pull/190/commits/ff934a8fbda9798699006a2aadf6471977843dee

Keeping it there for now but it needs to be tested in ORC. But making a separate PR for that, so in this PR keeping it explicit, because otherwise it defaults to --mm:orc yes.

websock.nimble
tersec: ok, nim 2.2.4 >= 2.0.16

arnetheduck (approved):

PR #194, opened 2026-05-06
feat: make rng agnostic

richard-ramos:
## Summary

Introduce a backend-neutral RNG abstraction for WebSocket randomness.

This PR adds `WebSocketRng`, backed by a random-bytes callback. BearSSL remains the default RNG and explicit BearSSL compatibility overloads are kept for `HmacDrbgContext` callers.

## Motivation

`nim-websock` currently exposes BearSSL RNG types in its public API. That makes downstream projects depend on BearSSL. This change lets callers provide randomness from any backend while preserving BearSSL compatibility

### DIFF
--- a/tests/testwebsockets.nim
+++ b/tests/testwebsockets.nim
@@ -19,6 +19,17 @@ import ./helpers
 
 let address = initTAddress("127.0.0.1:8888")
 
+suite "Test RNG":
+  test "custom random-bytes callback":
+    let rng = newWebSocketRng(
+      proc(dst: var openArray[byte]): bool {.closure, gcsafe, raises: [].} =
+        for i in 0 ..< dst.len:
+          dst[i] = byte(i + 1)
+        true
+    )
+
+    check MaskKey.random(rng) == [byte 1, 2, 3, 4]
+
 suite "Test handshake":
   setup:
     var
@@ -730,7 +741,7 @@ suite "Test Closing":
 
 suite "Test Payload":
   setup:
-    let rng {.used.} = HmacDrbgContext.new()
+    let rng {.used.} = newWebSocketRng()
     var
       server: HttpServer
 
@@ -834,7 +845,7 @@ suite "Test Payload":
       address = initTAddress("127.0.0.1:8888"),
       frameSize = maxFrameSize)
 
-    let maskKey = MaskKey.random(rng[])
+    let maskKey = MaskKey.random(rng)
     await session.stream.writer.write(
       (await Frame(
         fin: false,
@@ -894,7 +905,7 @@ suite "Test Payload":
         pong = true
     )
 
-    let maskKey = MaskKey.random(rng[])
+    let maskKey = MaskKey.random(rng)
     await session.stream.writer.write(
       (await Frame(
         fin: false,
--- a/tests/testwebsockets.nim
+++ b/tests/testwebsockets.nim
@@ -21,20 +21,18 @@ let address = initTAddress("127.0.0.1:8888")
 
 suite "Test RNG":
   test "custom random-bytes callback":
-    let rng = newWebSocketRng(
+    let rng: RandomBytesRng =
       proc(dst: var openArray[byte]): bool {.closure, gcsafe, raises: [].} =
         for i in 0 ..< dst.len:
           dst[i] = byte(i + 1)
         true
-    )
 
     check MaskKey.random(rng) == [byte 1, 2, 3, 4]
 
   test "custom random-bytes callback failure raises WSRngError":
-    let rng = newWebSocketRng(
-      proc(dst: var openArray[byte]): bool {.closure, gcsafe, raises: [].} =
+    let rng: RandomBytesRng =
+      proc(_: var openArray[byte]): bool {.closure, gcsafe, raises: [].} =
         false
-    )
 
     expect WSRngError:
       discard MaskKey.random(rng)
@@ -750,7 +748,7 @@ suite "Test Closing":
 
 suite "Test Payload":
   setup:
-    let rng {.used.} = newWebSocketRng()
+    let rng {.used.} = bearSslRng(HmacDrbgContext.new())
     var
       server: HttpServer
 
--- a/tests/testwebsockets.nim
+++ b/tests/testwebsockets.nim
@@ -30,6 +30,15 @@ suite "Test RNG":
 
     check MaskKey.random(rng) == [byte 1, 2, 3, 4]
 
+  test "custom random-bytes callback failure raises WSRngError":
+    let rng = newWebSocketRng(
+      proc(dst: var openArray[byte]): bool {.closure, gcsafe, raises: [].} =
+        false
+    )
+
+    expect WSRngError:
+      discard MaskKey.random(rng)
+
 suite "Test handshake":
   setup:
     var
--- a/websock/types.nim
+++ b/websock/types.nim
@@ -235,7 +235,8 @@ proc newWebSocketRng*(randomBytes: RandomBytesProc): WebSocketRng =
   WebSocketRng(randomBytes: randomBytes)
 
 proc generate*(rng: WebSocketRng, dst: var openArray[byte]): bool =
-  doAssert not rng.isNil, "rng cannot be null"
+  if rng.isNil or rng.randomBytes.isNil:
+    return false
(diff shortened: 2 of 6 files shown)


tests/testwebsockets.nim
```diff
@@ -19,6 +19,27 @@ import ./helpers
...
+          dst[i] = byte(i + 1)
+        true
+    )
+
+    check MaskKey.random(rng) == [byte 1, 2, 3, 4]
+
+  test "custom random-bytes callback failure raises WSRngError":
+    let rng = newWebSocketRng(
+      proc(dst: var openArray[byte]): bool {.closure, gcsafe, raises: [].} =
+        discard dst
```
tersec: ?

richard-ramos: just wanted to add some body to the callback, but never got anywhere... i'll just return false instead

tersec: If `dst` isn't going to be used, can use `_` instead:
- https://nim-lang.org/docs/manual.html#statements-and-expressions-special-identifier-nim-underscore
- https://nim-lang.org/docs/manual.html#statements-and-expressions-tuple-unpacking

This clarifies that it's not an oversight that `dst` isn't used and conveys an intentionality in a more direct way.

websock/types.nim
```diff
@@ -54,6 +54,11 @@ type
 
   MaskKey* = array[4, byte]
   WebSecKey* = array[16, byte]
+  RandomBytesProc* = proc(dst: var openArray[byte]): bool {.
+    closure, gcsafe, raises: []
+  .}
+  WebSocketRng* = ref object
```
arnetheduck: there's no point having a proc in a `ref` object like this - both are `ref` already so the added layer of heap allocation ends up not doing anything useful - we can use `type Rng = proc ...` and that's enough

arnetheduck: in other words, the closure environment of the proc already provides the advantages that another ref object wrapper could provide, itself being a ref object..

tests/testwebsockets.nim
richard-ramos: Implemented in https://github.com/status-im/nim-websock/pull/194/changes/424a2c53b245b0667e343cf0e9c0cb570c8d1bb9

websock/types.nim
richard-ramos: Implemented in https://github.com/status-im/nim-websock/pull/194/changes/424a2c53b245b0667e343cf0e9c0cb570c8d1bb9

websock/websock.nim
```diff
@@ -361,3 +434,22 @@ proc new*(
...
+
+proc new*(
+  _: typedesc[WSServer],
+  protos: openArray[string] = [""],
+  factories: openArray[ExtFactory] = [],
+  frameSize = WSDefaultFrameSize,
+  onPing: ControlCb = nil,
+  onPong: ControlCb = nil,
+  onClose: CloseCb = nil,
+  rng: ref HmacDrbgContext): WSServer =
```
tersec: Is the goal to `deprecate` this or keep it indefinitely as fully supported and documented part of the API?

This `HmacDrbgContext` thing is a bit of a wart, and even having the compatibility layer in `nim-websock` ultimately pushes this sort of random RNG source that happens to be a TLS library into websock's core dependencies.

tersec (approved):

websock/websock.nim
richard-ramos: Yes, I agree. I do not want `HmacDrbgContext` to be the main API.

IMO `nim-websock` only needs a way to “fill a buffer with random bytes”. It does not need to know if the source is BearSSL, BoringSSL, the OS, or something else. At the same time, it should still have a convenient default for users who do not want to think about rng setup.

The `HmacDrbgContext` overload can stay only as a compatibility wrapper for users who want to pass the BearSSL RNG directly. But IMO the long-term API should use `RandomBytesRng`, and this particular overload should be marked `.deprecated.`, with users wrapping the RNG explicitly using `bearSslRng` or maybe with a `converter` altho I see in the style guide that it's not recommended to use those.

wdyt?

websock/types.nim
```diff
@@ -54,6 +54,9 @@ type
 
   MaskKey* = array[4, byte]
   WebSecKey* = array[16, byte]
+  RandomBytesRng* = proc(dst: var openArray[byte]): bool {.
```
arnetheduck, line 57: `Rng` = random number generator so this name ends up a bit repetitive

websock/websock.nim
tersec: > At the same time, it should still have a convenient default for users who do not want to think about rng setup.

Except it doesn't really, currently. Passing the BearSSL RNG isn't not-thinking-about-RNG setup.

> The `HmacDrbgContext` overload can stay only as a compatibility wrapper for users who want to pass the BearSSL RNG directly.

Well this is part of the question/point, as long as this has to be supported, `nim-websock` (for no other reason I found) will pull in this separate TLS library, with its C build steps (Nim is bad at caching these, so they're slow and also relatively picky about system details so hinder portability) regardless of whether the user uses `HmacDrbgContext`.

Right now, IFT code does tend to use it, but alternatives exist, and `nim-websock` doesn't otherwise care. It's a fragile (requires various specific `config.nims` options to handle well in LTO, for example), heavyweight, and now-unmaintained dependency which calling code now can and should avoid.

So yes, it can't be removed immediately, but `{.deprecated.}` would at least mark the intent in a way which might open up removing this dependency.

arnetheduck: > for no other reason I found

chronos pulls in bearssl as well, unconditionally - a core feature of websock, like chronos, is to support TLS so until there's a different source of TLS (or this support is made optional), bearssl will be "mandatory" for the core feature set of these two libraries - so the rng by itself doesn't really change anything

tersec: By itself, no, and yes, chronos pulls it in. But `nim-websock` itself doesn't implement TLS or any cryptographic randomness TLS requires. It also does not appear to pass along its RNG to Chronos, though I might have missed it (I didn't scan thoroughly). But if it does, then this PR is ... kind of incomplete even strictly on the `nim-websock` side because it then needs that BearSSL PRNG for Chronos.

So addressing `nim-websock`, in its own direct needs, is a necessary step in proceeding.

Regardless, with QUIC, we'll need TLS 1.3 support, and BearSSL realistically will not within any reasonable timeframe reliably provide it. It's minimally maintained. Once we need that TLS 1.3 support, we need another TLS library, and at that point BearSSL is basically dead weight. We need all the (very unfortunate) build complications from BoringSSL/lsquic, etc, so BearSSL's own quirks add nothing. It's pulling in a whole TLS library for the PRNG. Extricating it would require changes in many places, yes, but that doesn't mean it's not worth doing in any one of them.

arnetheduck: > It's pulling in a whole TLS library for the PRNG. 

well, not quite - the main reason it's pulling in bearssl is because websock works over [tls](https://github.com/status-im/nim-websock/blob/02616eccd5c826b83c580fa092e9675da04491c7/websock/http/client.nim#L126) and chronos supplies tls via bearssl - practically, the rng does not add any (meaningful) compilation overhead on top of that (since both hmac and aes are used by tls internally - not through this rng however). 

As long as this is the case, this rng change doesn't have any practical impact apart from adding some purely theoretical abstraction for the sake of it. If we want to drop bearssl, the road begins with chronos and not this rng really - the rng is the smallest change at the tail end of the story.

> whole

The way the bearssl build works is that it only builds the parts that are actually used - ie using the RNG means only the rng part of bearssl will be built (which is a small subset - hmac/hash and aes - crucially, it doesn't pull in EC and math) .. if we were to use a different tls library but the same rng, you would see just the rng subset being compiled. 

True - if we switch to a different TLS library and that library also exposes a good RNG, we could switch the rng as well - not all TLS libs supply rngs though and we have traditionally not wanted to rely on the OS to provide randomness in large quantities since this falls apart on certain hardware - instead, we stretch a small os-provided seed using bearssl and its rng gen.

On a separate note - one of changes in nim 2.2.10 is that it _vastly_ improves compilation speed of C files (an order of magnitude for bearssl in particular) in general due to a [bugfix](https://github.com/nim-lang/Nim/commit/440b55a44a82fa4e6b1d17e36c46d517d2b06cab).

